PR #527, opened 2022-02-11
fix(query): validate scheduledSurfaceGuid on getScheduledCuratedCorpusItems

jpetto:
## Goal

fix a bug where an empty string could be passed in for `scheduledSurfaceGuid`, resulting in scheduled items being returned for _all_ scheduled surfaces.

## References

JIRA ticket:
* https://getpocket.atlassian.net/browse/BACK-1314


Herraj (approved):
💥

PR #528, opened 2022-02-14
feat(auth): Add access control checks to ScheduledItem mutations [BACK-1173]

nina-py:
## Goal

- Improved auth user context with more convenience props (heavily inspired by the approach taken in Prospect API).

- Added access control checks to `createScheduledItem` and `deleteScheduledItem` mutations.

- Updated existing tests. Added new integration tests that test various user privilege scenarios.

## References

JIRA ticket:
* https://getpocket.atlassian.net/browse/BACK-1173

### DIFF
--- a/src/admin/resolvers/mutations/RejectedItem.integration.ts
+++ b/src/admin/resolvers/mutations/RejectedItem.integration.ts
@@ -136,12 +136,10 @@ describe('mutations: RejectedItem', () => {
       expect(result.errors).not.to.be.null;
 
       // And there is the correct error from the resolvers
-      if (result.errors) {
-        expect(result.errors[0].message).to.contain(
-          `A rejected item with the URL "${input.url}" already exists.`
-        );
-        expect(result.errors[0].extensions?.code).to.equal('BAD_USER_INPUT');
-      }
+      expect(result.errors?.[0].message).to.contain(
+        `A rejected item with the URL "${input.url}" already exists.`
+      );
+      expect(result.errors?.[0].extensions?.code).to.equal('BAD_USER_INPUT');
 
       // Check that the REJECT_ITEM event was not fired
       expect(eventTracker.callCount).to.equal(0);
@@ -168,12 +166,10 @@ describe('mutations: RejectedItem', () => {
       expect(result.errors).not.to.be.null;
 
       // And there is the correct error from the resolvers
-      if (result.errors) {
-        expect(result.errors[0].message).to.contain(
-          `An approved item with the URL "${input.url}" already exists.`
-        );
-        expect(result.errors[0].extensions?.code).to.equal('BAD_USER_INPUT');
-      }
+      expect(result.errors?.[0].message).to.contain(
+        `An approved item with the URL "${input.url}" already exists.`
+      );
+      expect(result.errors?.[0].extensions?.code).to.equal('BAD_USER_INPUT');
 
       // Check that the REJECT_ITEM event was not fired
       expect(eventTracker.callCount).to.equal(0);
--- a/src/admin/resolvers/mutations/RejectedItem.integration.ts
+++ b/src/admin/resolvers/mutations/RejectedItem.integration.ts
@@ -1,19 +1,27 @@
 import { expect } from 'chai';
 import sinon from 'sinon';
-import { db, getServer } from '../../../test/admin-server';
+import { db } from '../../../test/admin-server';
 import {
   clearDb,
   createApprovedItemHelper,
   createRejectedCuratedCorpusItemHelper,
+  getServerWithMockedHeaders,
 } from '../../../test/helpers';
 import { CREATE_REJECTED_ITEM } from './sample-mutations.gql';
 import { CreateRejectedItemInput } from '../../../database/types';
 import { CuratedCorpusEventEmitter } from '../../../events/curatedCorpusEventEmitter';
 import { ReviewedCorpusItemEventType } from '../../../events/types';
+import { MozillaAccessGroup } from '../../../shared/types';
 
 describe('mutations: RejectedItem', () => {
   const eventEmitter = new CuratedCorpusEventEmitter();
-  const server = getServer(eventEmitter);
+
+  const headers = {
+    name: 'Test User',
+    username: 'test.user@test.com',
+    groups: `group1,group2,${MozillaAccessGroup.SCHEDULED_SURFACE_CURATOR_FULL}`,
+  };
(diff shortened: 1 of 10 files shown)


src/admin/resolvers/mutations/RejectedItem.integration.ts
```diff
@@ -1,19 +1,27 @@
...
 describe('mutations: RejectedItem', () => {
   const eventEmitter = new CuratedCorpusEventEmitter();
-  const server = getServer(eventEmitter);
+
+  const headers = {
+    name: 'Test User',
+    username: 'test.user@test.com',
+    groups: `group1,group2,${MozillaAccessGroup.SCHEDULED_SURFACE_CURATOR_FULL}`,
+  };
+  const server = getServerWithMockedHeaders(headers, eventEmitter);
```
nina-py, line 24: As an aside, prepped context/mock headers for other mutation-related integration tests.

src/admin/resolvers/mutations/ScheduledItem.integration.ts
```diff
@@ -217,6 +225,116 @@ describe('mutations: ScheduledItem', () => {
...
+      // ...without success. There is no data
+      expect(result.data).to.be.null;
+
+      expect(result.errors).not.to.be.null;
+
+      if (result.errors) {
+        // And there is an access denied error
+        expect(result.errors[0].message).to.contain(ACCESS_DENIED_ERROR);
+      }
+
```
Herraj, line 301: I just learned from Kelvin the other day that you have to do this to be able to do optional chaining on an possibly undefined array 🙃 :

`result.errors?.[0].message` have to add the `.` before we can access the array index, which is absolutely not how you access array indices. With this you can get rid of the `if()`

src/admin/resolvers/mutations/ScheduledItem.integration.ts
```diff
@@ -324,5 +442,112 @@ describe('mutations: ScheduledItem', () => {
         await eventTracker.getCall(0).args[0].scheduledCorpusItem.externalId
       ).to.equal(scheduledItem.externalId);
     });
+
+    it('should fail if user has read-only access', async () => {
+      const headers = {
+        name: 'Test User',
+        username: 'test.user@test.com',
+        groups: `group1,group2,no-access-for-you`,
```
Herraj, line 450: Same comment as above regarding the read only test

src/admin/resolvers/mutations/ScheduledItem.ts
```diff
@@ -3,23 +3,41 @@ import {
...
  */
 export async function deleteScheduledItem(
   parent,
   { data },
-  context
+  context: IContext
 ): Promise<ScheduledItem> {
+  // Need to fetch the item first to check access privileges.
+  // Could we make it easier/not hit the DB by passing the Scheduled Surface GUID
+  // as a required input to the delete mutation?
```
Herraj, line 30: I ran into this as well when updating a prospect as curated. The request only sends `prospectId` so I had to do a DB fetch first to get the `guid` on it.

@jpetto and I had a discussion about it whether the FE should pass in the `guid` as well but I think we should be okay making this extra DB call given we will only have a handful of users using this (for now 😄 )

src/admin/resolvers/mutations/ScheduledItem.ts
```diff
@@ -3,23 +3,41 @@ import {
...
+  const item = await context.db.scheduledItem.findUnique({
+    where: { externalId: data.externalId },
+  });
+
+  // Check if the user can execute this mutation.
+  if (item && !context.authenticatedUser.canWrite(item.scheduledSurfaceGuid)) {
+    throw new AuthenticationError(ACCESS_DENIED_ERROR);
+  }
```
Herraj, line 38: should we check first if `item` exists or does the `findUnique()` throw an error if it doesn't find an entry?

The reason I ask is if we have this scenario where item is not found and user _**is**_ authenticated to write, the if condition would be: 

`if( false && !True) --> if(false)` it would skip this if block and  continue to delete it

src/shared/types.ts
```diff
@@ -127,3 +137,6 @@ export enum Topics {
   TECHNOLOGY = 'TECHNOLOGY',
   TRAVEL = 'TRAVEL',
 }
+
+export const ACCESS_DENIED_ERROR =
+  'You do not have access to perform this action.';
```
Herraj, line 142: Nice, this is a very good practice!

Herraj (commented):
Good work! Just a few comments!

src/admin/resolvers/mutations/ScheduledItem.integration.ts
```diff
@@ -217,6 +225,116 @@ describe('mutations: ScheduledItem', () => {
...
```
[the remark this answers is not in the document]
nina-py: Good point! :)

src/admin/resolvers/mutations/ScheduledItem.integration.ts
nina-py: Thanks for the tip, @Herraj! There were quite a few instances of the `if()` statement being used around these checks in integration tests so I went and updated them all.

src/admin/resolvers/mutations/ScheduledItem.integration.ts
nina-py: Good catch, thank you!

src/admin/resolvers/mutations/ScheduledItem.ts
nina-py: Thank you Herraj, updated the comment after our discussion noting that it's ok to leave that extra DB call in place.

src/admin/resolvers/mutations/ScheduledItem.ts
nina-py: @Herraj , the `dbDeleteScheduledItem()` function called immediately after throwing the authentication error will actually throw a Not Found error if the scheduled item is nowhere to be found in the database. 

We could potentially move that check right into the resolver and place it earlier in the chain, before the auth check, but this becomes almost a philosophical question - do we want to let unauthorised users know if something is not in the database? Or do we want to throw the authorisation error first if their privileges are not sufficient? @jpetto, any thoughts?

Herraj (approved):
💥

src/admin/resolvers/mutations/ScheduledItem.ts
jpetto: we should throw authorization errors before any others. it's an internal tool so the stakes are low, but i'd rather keep good security practices 😄

PR #534, opened 2022-02-15
feat(auth): [BACK-1175] add auth to reject item mutation

Herraj:
## Goal
Add an auth check to the `rejectApprovedItem` and `createRejectedApprovedItemMutation` mutation. 

**Aside**: this also includes the docker support for arm64

JIRA ticket:
* [BACK-1175]

[BACK-1175]: https://getpocket.atlassian.net/browse/BACK-1175?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/admin/resolvers/mutations/RejectedItem.ts
```diff
@@ -13,8 +16,13 @@ import { ReviewedCorpusItemEventType } from '../../../events/types';
 export async function createRejectedItem(
   parent,
   { data },
-  context
+  context: IContext
 ): Promise<RejectedCuratedCorpusItem> {
+  // check if user is not authorized to reject an item
+  if (!context.authenticatedUser.hasFullAccess) {
```
jpetto, line 22: this might get complicated. we have two scenarios for rejection:

1. during prospecting
2. from the corpus

these have pretty different use cases, and as `scheduledSurfaceGuid` isn't coming in on this mutation, we have some decisions to make. taking this to slack.

jpetto: okay based on carolyn's answer in slack, this should be updated to allow anyone with write access to perform rejections (access to any scheduled surface OR full access).

jpetto (changes requested):
code as-is is great - unfortunately more requirements gathering happened and we'll need to update the auth check and add some test scenarios around it. hopefully straight forward!

src/admin/resolvers/mutations/RejectedItem.ts
nina-py: @Herraj , @jpetto , I've split up the `canWrite()` convenience method in the context into `canWriteToSurface()` (membership of full access group or access to _this_ particular surface is accepted) and `canWriteToCorpus()` (membership of full access group or access to _any_ scheduled surface is accepted) in https://github.com/Pocket/curated-corpus-api/pull/536. If you like this pattern it could be used here?

jpetto (approved):
nice! 🚀

PR #536, opened 2022-02-16
feat(auth): Update `updateApprovedCuratedCorpusItem` mutation to check access [BACK-1320]

nina-py:
## Goal

_What changed? What is the business/product goal?_

We now have access checks in the `updateApprovedCuratedCorpusItem` mutation! 🎉 

- Updated convenience methods in context to include `canWriteToCorpus` and `canWriteToSurface`.

- Added an access check to the `updateApprovedCuratedCorpusItem` mutation.

- Added integration tests.

## References

JIRA ticket:
* https://getpocket.atlassian.net/browse/BACK-1320


src/admin/context.ts
```diff
@@ -75,29 +76,61 @@ export class ContextManager implements IContext {
...
+          MozillaAccessGroup.NEW_TAB_CURATOR_ENUS as string, // Oh, TypeScript!
+          MozillaAccessGroup.NEW_TAB_CURATOR_ENGB as string,
+          MozillaAccessGroup.NEW_TAB_CURATOR_DEDE as string,
+          MozillaAccessGroup.NEW_TAB_CURATOR_ENINTL as string,
+          MozillaAccessGroup.POCKET_HITS_CURATOR_ENUS as string,
+          MozillaAccessGroup.POCKET_HITS_CURATOR_DEDE as string,
+        ];
+
+        return (
+          accessGroups.filter((element) => {
```
kschelonka, line 118: consider set intersection as a simpler alternative

kschelonka: uhh, nvm, javascript does not have simple set intersection???????

nina-py: Exactly, Kat!

jpetto: we could maybe use `filter` here? but that's a nit-pick and doesn't block.

jpetto (approved):
nice! love the refactors, too.

PR #540, opened 2022-02-17
feat(auth): Add access checks around creating a corpus item [BACK-1325]

nina-py:
## Goal

- [x] Add an access check to creating a corpus item (done, anyone who has write access can do that).
- [x] Add integration tests for the above
- [x] Get one bizarre test failure to go away (?!!) - fixed by splitting auth-related tests into their own `describe()` block.
- [x] Add an extra access check in the same mutation for the optional scheduling - the solution we arrived to while pairing with @jpetto is to check for `canWriteToSurface` access if scheduling is specified and to check for `canWriteToCorpus` if item is only being added to the corpus.
- [x] Add more integration tests

## References

JIRA ticket:
* https://getpocket.atlassian.net/browse/BACK-1325

### DIFF
--- a/src/admin/resolvers/mutations/ApprovedItem.integration.ts
+++ b/src/admin/resolvers/mutations/ApprovedItem.integration.ts
@@ -71,7 +71,7 @@ describe('mutations: ApprovedItem', () => {
       isSyndicated: false,
     };
 
-    it('creates an approved item with all inputs supplied', async () => {
+    it('should create an approved item if user has full access', async () => {
       // Set up event tracking
       const eventTracker = sinon.fake();
       eventEmitter.on(ReviewedCorpusItemEventType.ADD_ITEM, eventTracker);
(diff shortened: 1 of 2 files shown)


src/admin/resolvers/mutations/ApprovedItem.integration.ts
```diff
@@ -604,3 +604,136 @@ describe('mutations: ApprovedItem', () => {
...
+    language: 'de',
+    publisher: 'Convective Cloud',
+    topic: 'Technology',
+    isCollection: false,
+    isTimeSensitive: true,
+    isSyndicated: false,
+  };
+
+  describe('createApprovedCuratedCorpusItem mutation', () => {
+    it('should succeed if user has access to one of scheduled surfaces', async () => {
```
Herraj, line 628: Just making sure, so a user can create an approved item as long as they have _at least one_ scheduled surface access?

Herraj (approved):
Good stuff! 🚀

PR #543, opened 2022-02-18
feat(auth): update `createdBy`/`updatedBy` fields with SSO user info [BACK-1326]

nina-py:
## Goal

- Updated `createdBy`/`updatedBy` fields in all relevant mutations(one notable exception is the image upload mutation) to record the SSO username.

- Updated integration tests with assertions expecting to see the SSO username returned for mutated entities.

- One important thing discussed in the ticket (as of writing this, it's still undecided) is whether to use the username at all, and if yes, the whole thing returned by Cognito or a portion of it - e.g., the username.

## References

JIRA ticket:
* https://getpocket.atlassian.net/browse/BACK-1326


nina-py:
@Herraj, the ticket has some discussion around this, and also whether to use the username or the full name of the user. I'm following through with the default username and then we can tweak it afterwards.

src/admin/resolvers/mutations/ScheduledItem.ts
```diff
@@ -43,6 +43,13 @@ export async function deleteScheduledItem(
   // Access allowed, proceed as normal from this point on.
   const scheduledItem = await dbDeleteScheduledItem(context.db, data);
 
+  // Before we send the event to Snowplow, update the `updatedBy` and `updatedAt` fields
+  // as the object returned from the database resolver will have the details
+  // of the previous update and not the final one (aka the hard delete).
+  scheduledItem.updatedBy = context.authenticatedUser.username;
+  // Do we need the date in UTC here?
```
jpetto, line 50: the snowplow schema seems to call for a UTC unix timestamp, so i think we do need to [convert this](https://stackoverflow.com/a/11893157).

jpetto (changes requested):
only issue i see here is the date format for the snowplow event. otherwise looks good!

looks like we landed on a different value for `username` in the ticket, but this will be updated at the gateway level and doesn't need to block this PR. (i will file a ticket for the gateway work shortly.)

src/admin/resolvers/mutations/ScheduledItem.ts
nina-py: @jpetto, I've double-checked here and updated the comment to reflect my findings - we do convert the date to a Unix timestamp in the `*SnowplowHandler` classes, so there is no need to do it there. Prisma also returns a JS `Date` object so it's more consistent if we set it to just `Date()` here.

Herraj (approved):
Good stuff! :shipit:

PR #560, opened 2022-02-24
feat(auth): Add auth check to `uploadApprovedItemImage` mutation [BACK-1346]

nina-py:
## Goal

This is fairly straightforward. As long as the user has write access to the corpus, they should be able to upload images to S3. 

- Added an authentication check to the mutation.

- Added an integration test.

## References

JIRA ticket: https://getpocket.atlassian.net/browse/BACK-1346


Herraj (approved):
🚀

PR #565, opened 2022-02-25
fix(validation): validate topic and rejection reason enums on mutations

jpetto:
## Goal

prevent invalid `Topic` and `RejectionReason` values from making it into the db.

- not-very-DRY code. should these checks go into a helper?

## References

JIRA ticket:
* https://getpocket.atlassian.net/browse/BACK-1351

### DIFF
--- a/src/admin/resolvers/mutations/RejectedItem.ts
+++ b/src/admin/resolvers/mutations/RejectedItem.ts
@@ -1,8 +1,9 @@
+import { UserInputError } from 'apollo-server';
 import { RejectedCuratedCorpusItem } from '@prisma/client';
 import { AuthenticationError } from 'apollo-server-core';
 import { createRejectedItem as dbCreateRejectedItem } from '../../../database/mutations';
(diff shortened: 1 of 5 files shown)


src/admin/resolvers/mutations/RejectedItem.ts
```diff
@@ -23,6 +24,20 @@ export async function createRejectedItem(
     throw new AuthenticationError(ACCESS_DENIED_ERROR);
   }
 
+  // validate reason enum
+
+  // rejection reason comes in as a comma separated string
+  // split them up and strip out preceeding/trailing white space
+  const reasons = data.reason.split(',').map((reason) => {
+    return reason.trim();
```
kschelonka, line 32: if you wanted to simplify a bit, you could move the for loop logic into this iterator (either using forEach, or map, whichever)

kschelonka (approved):

Herraj (approved):
:shipit:

PR #567, opened 2022-02-25
feat(auth): [BACK-1345] add auth to `getApprovedCuratedCorpusItemByUrl` query

Herraj:
## Goal
Added authentication to `getApprovedCuratedCorpusItemByUrl` query

Moved authentication tests to a separate file

JIRA ticket:
* [BACK-1345]

[BACK-1345]: https://getpocket.atlassian.net/browse/BACK-1345?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/admin/resolvers/queries/ApprovedItem.auth.integration.ts
+++ b/src/admin/resolvers/queries/ApprovedItem.auth.integration.ts
@@ -16,6 +16,34 @@ describe('queries: ApprovedCuratedCorpusItem - authentication', () => {
   beforeAll(async () => {
     // clear out db to start fresh
     await clearDb(db);
+
+    // Create some items
+    const stories = [
+      {
+        title: 'How To Win Friends And Influence People with GraphQL',
+        language: 'en',
+        status: CuratedStatus.RECOMMENDATION,
+        topic: 'FOOD',
+      },
+      {
+        title: 'What Zombies Can Teach You About GraphQL',
+        language: 'en',
+        status: CuratedStatus.RECOMMENDATION,
+        url: 'https://www.sample-domain/what-zombies-can-teach-you-graphql',
+        topic: 'TECHNOLOGY',
+      },
+      {
+        title: 'How To Make Your Product Stand Out With GraphQL',
+        language: 'en',
+        status: CuratedStatus.RECOMMENDATION,
+        topic: 'TECHNOLOGY',
+      },
+    ];
+
+    // insert test stories into the db
+    for (const story of stories) {
+      await createApprovedItemHelper(db, story);
+    }
   });
 
   afterAll(async () => {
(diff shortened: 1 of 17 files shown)


src/admin/resolvers/queries/ApprovedItem.auth.integration.ts
```diff
@@ -0,0 +1,264 @@
...
+      expect(result.errors).not.to.be.undefined;
+
+      // check if the error we get is access denied error
+      expect(result.errors?.[0].message).to.equal(ACCESS_DENIED_ERROR);
+
+      await server.stop();
+    });
+  });
+
+  describe('getApprovedCuratedCorpusItemByUrl query', () => {
```
Herraj, line 161: Tests for this query start from here. The ones above are simply moved to this file.

src/admin/resolvers/queries/ApprovedItem.auth.integration.ts
```diff
@@ -0,0 +1,264 @@
...
+        query: GET_APPROVED_ITEM_BY_URL,
+        variables: {
+          url: 'https://www.test.com/story-two',
+        },
+      });
+
+      expect(result.data).not.to.be.null;
+      expect(result.errors).to.be.undefined;
+
+      await server.stop();
```
kkyeboah, line 183: could move to `afterAll`?

src/admin/resolvers/queries/ApprovedItem.auth.integration.ts
```diff
@@ -0,0 +1,264 @@
...
+      await server.start();
+
+      const result = await server.executeOperation({
+        query: GET_APPROVED_ITEM_BY_URL,
+        variables: {
+          url: 'https://www.test.com/story-two',
+        },
+      });
+
+      expect(result.data).not.to.be.null;
```
kkyeboah, line 204: to make this test a little better, we should have at least 2 records, fetch all and ensure that only ones that can be access by this user is returned

src/admin/resolvers/queries/ApprovedItem.auth.integration.ts
```diff
@@ -0,0 +1,264 @@
...
+
+      // insert test stories into the db
+      for (const story of stories) {
+        await createApprovedItemHelper(db, story);
+      }
+    });
+
+    afterAll(async () => {
+      await db.$disconnect();
+    });
```
nina-py, line 58: I don't think this second DB disconnect should be in this file at all - the one within the top-level `describe()` block is probably enough.

src/admin/resolvers/queries/ApprovedItem.auth.integration.ts
```diff
@@ -0,0 +1,264 @@
...
+          status: CuratedStatus.RECOMMENDATION,
+          topic: 'TECHNOLOGY',
+        },
+      ];
+
+      // insert test stories into the db
+      for (const story of stories) {
+        await createApprovedItemHelper(db, story);
+      }
+    });
```
nina-py, line 54: It looks like the newly added files (for the `getApprovedCuratedCorpusItemByUrl` query) are relying on the data seeded in another `describe()` block - not the top-level one but a block on the same level that tests auth checks for `getApprovedCuratedCorpusItems`. 

Would it make sense to move it up to the top-level `describe() -> beforeAll()` so that when we come back to these tests in a while there is no confusion about where the data for the first query is coming from?

src/admin/resolvers/queries/ApprovedItem.auth.integration.ts
Herraj: That's a good idea but for this query, the way the auth works is that if you have access to any one surface, you will get full access. So I will be checking for all the items since all of them will be returned.

Also the reason I omitted checking specific items for this test is because this focused on auth. The other tests in the `ApprovedItem.integration` file cover that part. Omitting the insertion of items in the DB also makes this slightly faster and kinda less redundant in a way 😄

src/admin/resolvers/queries/ApprovedItem.auth.integration.ts
Herraj: I am instantiating a new `server` object in each test (`it()`) so I can't really do it in `afterAll()`

These tests will be refactored to avoid using a new `server` object. Once that is done, we'll just have one `server` in the test set up calls!

jpetto (approved):
nice cleanup 😄

jpetto (approved):

PR #570, opened 2022-02-28
feat(auth): [BACK-1343] Add auth to `getRejectedCuratedCorpusItems` query

nina-py:
## Goal

- Added authentication checks to `getRejectedCuratedCorpusItems` query.

- Added auth-related integration tests.

- Added mock headers to existing tests for Rejected Item queries so that they keep passing after the change.

## References

JIRA ticket:
* https://getpocket.atlassian.net/browse/BACK-1343

### DIFF
--- a/src/admin/resolvers/queries/RejectedItem.auth.integration.ts
+++ b/src/admin/resolvers/queries/RejectedItem.auth.integration.ts
@@ -1,0 +1,145 @@
+import { db } from '../../../test/admin-server';
+import {
+  clearDb,
+  createRejectedCuratedCorpusItemHelper,
+  getServerWithMockedHeaders,
+} from '../../../test/helpers';
+import { ACCESS_DENIED_ERROR, MozillaAccessGroup } from '../../../shared/types';
+import { GET_REJECTED_ITEMS } from './sample-queries.gql';
+import { expect } from 'chai';
+
+describe('queries: RejectedCuratedCorpusItem (authentication)', () => {
+  // A few sample Rejected Curated Corpus items - we don't need a lot of these
+  // for auth checks
+  const rejectedCuratedCorpusItems = [
+    {
+      title: '10 Unforgivable Sins Of PHP',
+    },
+    {
+      title: 'Take The Stress Out Of PHP',
+    },
+    {
+      title: 'The Untold Secret To Mastering PHP In Just 3 Days',
+    },
+    {
+      title: 'You Can Thank Us Later - 3 Reasons To Stop Thinking About PHP',
+    },
+  ];
+
+  let headers: {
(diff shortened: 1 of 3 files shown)


src/admin/resolvers/queries/RejectedItem.auth.integration.ts
```diff
@@ -0,0 +1,145 @@
...
+    {
+      title: 'Take The Stress Out Of PHP',
+    },
+    {
+      title: 'The Untold Secret To Mastering PHP In Just 3 Days',
+    },
+    {
+      title: 'You Can Thank Us Later - 3 Reasons To Stop Thinking About PHP',
+    },
+  ];
```
nina-py, line 27: Added only a handful of rejected corpus items for this suite of tests as we don't need many to make sure users get to see all of the Rejected Corpus if they are entitled to do so, and we don't need to run queries narrowed down by language or any other fields or queries limited by pagination - all of this is tested thoroughly in existing integration tests.

Also, following an emerging pattern from https://github.com/Pocket/curated-corpus-api/pull/567 by placing auth-related integration tests into its own file.

src/admin/resolvers/queries/RejectedItem.auth.integration.ts
```diff
@@ -0,0 +1,145 @@
...
+      const result = await server.executeOperation({
+        query: GET_REJECTED_ITEMS,
+      });
+
+      expect(result.data).to.be.null;
+      expect(result.errors).not.to.be.undefined;
+
+      // check if the error we get is the access denied error
+      expect(result.errors?.[0].message).to.equal(ACCESS_DENIED_ERROR);
+      expect(result.errors?.[0].extensions?.code).to.equal('UNAUTHENTICATED');
```
nina-py, line 120: Trying to check for the error code where possible. It's possible here but I remember we had issues before with other error codes returned by `apollo-server-errors`.

src/admin/resolvers/queries/RejectedItem.auth.integration.ts
```diff
@@ -0,0 +1,145 @@
...
+      );
+
+      await server.stop();
+    });
+
+    it('should throw an error when user does not have the required access', async () => {
+      // Set up auth headers with no valid access group
+      headers = {
+        ...headers,
+        groups: `this,that`,
```
Herraj, line 106: 😄

src/admin/resolvers/queries/RejectedItem.auth.integration.ts
Herraj: This is nice, let me do that in my tests too

Herraj (approved):
Nice! :shipit:

PR #577, opened 2022-03-01
feat(mutation): add mutation for rescheduling a scheduled item

jpetto:
## Goal

add a mutation for rescheduling a scheduled item

- remove superfluous server start/stop calls in test file
- add a new snowplow type for reschedule
- refactor delete scheduled item a bit to be more effecient

## References

JIRA ticket:
* https://getpocket.atlassian.net/browse/BACK-1354

### DIFF
--- a/src/admin/resolvers/mutations/ScheduledItem.ts
+++ b/src/admin/resolvers/mutations/ScheduledItem.ts
@@ -41,10 +41,7 @@ export async function deleteScheduledItem(
   }
 
   // Check if the user can execute this mutation.
-  if (
-    item &&
-    !context.authenticatedUser.canWriteToSurface(item.scheduledSurfaceGuid)
-  ) {
+  if (!context.authenticatedUser.canWriteToSurface(item.scheduledSurfaceGuid)) {
     throw new AuthenticationError(ACCESS_DENIED_ERROR);
   }
 
@@ -145,10 +142,7 @@ export async function rescheduleScheduledItem(
   }
 
   // Check if the user can execute this mutation.
-  if (
-    item &&
-    !context.authenticatedUser.canWriteToSurface(item.scheduledSurfaceGuid)
-  ) {
+  if (!context.authenticatedUser.canWriteToSurface(item.scheduledSurfaceGuid)) {
     throw new AuthenticationError(ACCESS_DENIED_ERROR);
   }
 
--- a/src/admin/resolvers/mutations/ScheduledItem.ts
+++ b/src/admin/resolvers/mutations/ScheduledItem.ts
@@ -1,6 +1,7 @@
 import {
   deleteScheduledItem as dbDeleteScheduledItem,
   createScheduledItem as dbCreateScheduledItem,
+  rescheduleScheduledItem as dbRescheduleScheduledItem,
 } from '../../../database/mutations';
 import { ScheduledItem } from '../../../database/types';
 import {
@@ -11,6 +12,7 @@ import { ScheduledCorpusItemEventType } from '../../../events/types';
 import { UserInputError } from 'apollo-server';
 import { PrismaClientKnownRequestError } from '@prisma/client/runtime';
 import { AuthenticationError } from 'apollo-server-errors';
+import { NotFoundError } from '@pocket-tools/apollo-utils';
 import { IContext } from '../../context';
 
 /**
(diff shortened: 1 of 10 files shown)


src/database/mutations/ScheduledItem.ts
```diff
@@ -53,26 +54,30 @@ export async function deleteScheduledItem(
...
+  db: PrismaClient,
+  data: RescheduleScheduledItemInput,
+  username: string
+): Promise<ScheduledItem> {
+  return await db.scheduledItem.update({
+    where: { externalId: data.externalId },
+    data: {
+      scheduledDate: data.scheduledDate,
+      updatedBy: username,
+      updatedAt: new Date(),
```
jpetto, line 77: is this the right format?

src/admin/resolvers/mutations/ScheduledItem.ts
```diff
@@ -117,3 +125,64 @@ export async function createScheduledItem(
...
+
+  if (!item) {
+    throw new NotFoundError(
+      `Item with ID of '${data.externalId}' could not be found.`
+    );
+  }
+
+  // Check if the user can execute this mutation.
+  if (
+    item &&
```
bassrock, line 149: should be able to remove this first conditional

src/database/mutations/ScheduledItem.ts
```diff
@@ -53,26 +54,30 @@ export async function deleteScheduledItem(
   db: PrismaClient,
   data: DeleteScheduledItemInput
 ): Promise<ScheduledItem> {
-  // Get the item to return with the mutation
-  const scheduledItem = await db.scheduledItem.findUnique({
-    where: { externalId: data.externalId },
+  return await db.scheduledItem.delete({
```
bassrock, line 57: i like this change

src/admin/resolvers/mutations/ScheduledItem.ts
```diff
@@ -32,6 +34,12 @@ export async function deleteScheduledItem(
...
 
+  if (!item) {
+    throw new NotFoundError(
+      `Item with ID of '${data.externalId}' could not be found.`
+    );
+  }
+
   // Check if the user can execute this mutation.
   if (
     item &&
```
bassrock, line 45: Could remove this conditional

src/admin/resolvers/mutations/ScheduledItem.ts
```diff
@@ -117,3 +125,64 @@ export async function createScheduledItem(
...
+  if (!item) {
+    throw new NotFoundError(
+      `Item with ID of '${data.externalId}' could not be found.`
+    );
+  }
+
+  // Check if the user can execute this mutation.
+  if (
+    item &&
+    !context.authenticatedUser.canWriteToSurface(item.scheduledSurfaceGuid)
```
bassrock, line 150: i really like how auth happened in these repos. this is very nice.

bassrock (commented):
Will let someone more knowledagebale then me 👍 it but had a few comments

src/database/mutations/ScheduledItem.ts
jpetto: me too!

kkyeboah (commented):
same minor comments as @bassrock about removing `item` checks from auth checks.

src/database/mutations/ScheduledItem.ts
nina-py: @jpetto, I don't think you need to set `updatedAt` manually at all, MySQL will update it for you automatically.

jpetto: oh nice! will verify.

jpetto: yes indeed 😄 

https://www.prisma.io/docs/reference/api-reference/prisma-schema-reference#updatedat

src/admin/resolvers/mutations/ScheduledItem.ts
```diff
@@ -117,3 +122,61 @@ export async function createScheduledItem(
...
+  } catch (error) {
+    // If it's the duplicate scheduling constraint, catch the error
+    // and send a user-friendly one to the client instead.
+    if (
+      error instanceof PrismaClientKnownRequestError &&
+      error.code === 'P2002'
+    ) {
+      throw new UserInputError(
+        `This story is already scheduled to appear on ${data.scheduledDate.toLocaleString(
+          'en-US',
```
Herraj, line 171: Wondering if the `en-US` date format would be confusing for the non North American curators?

Herraj (approved):
Good stuff! Assuming tests will come in another PR?

src/admin/resolvers/mutations/ScheduledItem.ts
jpetto: tests are in there! github just collapses the file because it's a big one. 😄 

as for the date format - maybe. (for better or for worse) our entire interface is US-centric. maybe down the road we can figure out how to display the date based on the current user's locale, but today is not that day.

PR #595, opened 2022-03-09
feat(graphql): [INFRA-330] Schema proposal for `importApprovedCuratedCorpusItem` mutation

nina-py:
## Goal

Propose a schema for a new mutation to be used by an automated process (immediate use case: curation tools data migration, potential future use: ML, etc.) to add an approved curated item to the corpus.

- The new mutation exposes extra inputs for an automated process to use: `createdBy`, `createdAt`, `updatedBy`. `isCollection`, `isSyndicated` and `isTimeSensitive` inputs are optional since these properties don't exist in the database we're migrating from.

- The new mutation returns a payload with both the approved item (always guaranteed) and the optional scheduled item if the story being added is scheduled onto a surface at the same time.

- Naming the mutation itself. Draft name is `loadApprovedCuratedCorpusItem`. Update: settled on `importApprovedCuratedCorpusItem`. 
- Whether we should keep all the default inputs carried over from `createApprovedCuratedCorpusItem` to future-proof this or cut them out to cater for just the curation tools data migration.

## References

JIRA ticket:
* https://getpocket.atlassian.net/browse/INFRA-330


schema-admin.graphql
```diff
@@ -487,6 +487,97 @@ input CreateApprovedCuratedCorpusItemInput {
...
+    """
+    isSyndicated: Boolean
+    """
+    A Unix timestamp of when the entity was created.
+    """
+    createdAt: Int!
+    """
+    A single sign-on user identifier of the user who created this entity.
+    """
+    createdBy: String!
```
Herraj, line 552: Maybe I missed it but do we have a `createdBy` for the entries/items that will be migrated? 

cc @sri-kirsh

Herraj (commented):
Looks good! Just one clarification comment :)

I'm okay with using the `load` prefix. Although, I do like `InsertApprovedCuratedCorpusItem` as well

schema-admin.graphql
```diff
@@ -487,6 +487,97 @@ input CreateApprovedCuratedCorpusItemInput {
...
 
+"""
+Input data for loading an Approved Item via an automated process and optionally scheduling
+this item to appear on a Scheduled Surface.
+"""
+input LoadApprovedCuratedCorpusItemInput {
+    """
+    The GUID of the corresponding Prospect ID. Will be empty for manually added items.
+    """
+    prospectId: ID
```
sri-kirsh, line 498: the prospectId will be passed as `null` for these mutation, so do we need this in the input?

Herraj: I think it should be okay since it's not a required field, we can pass it as `null`

schema-admin.graphql
```diff
@@ -487,6 +487,97 @@ input CreateApprovedCuratedCorpusItemInput {
...
+    publisher: String!
+    """
+    The image URL for this item's accompanying picture.
+    """
+    imageUrl: Url!
+    """
+    A topic this story best fits in.
+    Temporarily a string value that will be provided by Prospect API, possibly an enum in the future.
+    """
+    topic: String!
```
sri-kirsh, line 531: just calling out, for some items - the `topics` are null. we will pass them as `imported`

schema-admin.graphql
sri-kirsh: we have `createdBy` field but they are not always SSO username

schema-admin.graphql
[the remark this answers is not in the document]
jpetto: after some [conversation in slack](https://pocket.slack.com/archives/CTLKW5WCF/p1646842406714539), it has been decided that we will not be importing `prospectId`, so we can get rid of this input property.

schema-admin.graphql
```diff
@@ -487,6 +487,97 @@ input CreateApprovedCuratedCorpusItemInput {
     scheduledSurfaceGuid: ID
 }
 
+"""
+Input data for loading an Approved Item via an automated process and optionally scheduling
+this item to appear on a Scheduled Surface.
+"""
+input LoadApprovedCuratedCorpusItemInput {
```
sri-kirsh, line 494: we need a `source` field as input  - can add them as enum 

may be `BACKFILL` and `CLIENT`

schema-admin.graphql
```diff
@@ -674,6 +765,14 @@ type Mutation {
         data: CreateApprovedCuratedCorpusItemInput!
     ): ApprovedCuratedCorpusItem!
 
+    """
+    Lets an automated process create an Approved Item and optionally schedule it to appear
+    on a Scheduled Surface.
+    """
+    loadApprovedCuratedCorpusItem(
```
jpetto, line 772: i don't love "load" - it could be interpreted as "load from this system" as opposed to "load from another system". what about "import"?

jpetto (changes requested):

schema-admin.graphql
sri-kirsh: update, we have only 10 curators and only few of them are non-sso names. we can individually map them 
https://pocket.slack.com/archives/CTLKW5WCF/p1646848666279309?thread_ts=1646847427.826849&cid=CTLKW5WCF

schema-admin.graphql
sri-kirsh: created a follow up ticket to figure out this use case: https://getpocket.atlassian.net/browse/INFRA-376

schema-admin.graphql
```diff
@@ -487,6 +487,97 @@ input CreateApprovedCuratedCorpusItemInput {
...
+    """
+    language: String!
+    """
+    The name of the online publication that published this story.
+    """
+    publisher: String!
+    """
+    The image URL for this item's accompanying picture.
+    """
+    imageUrl: Url!
```
sri-kirsh, line 526: question: if we pass the `http://image_src` url, can the curatedCorpusApi fetch the byte stream and create a s3Url and return back with the response? 

This way, we can reduce two mutation calls to one and the server can take care of doing the node fetch to get the bytestream from the url

schema-admin.graphql
nina-py: Sorry could you clarify @sri-kirsh? Will topics be sent to Curated Corpus API as `null` or as a string value of `imported`?

schema-admin.graphql
nina-py: I love `import`, @jpetto , thank you for this suggestion! This also means the terribly named `LoadApprovedCuratedItemPayload` can become `ImportApprovedCuratedItemPayload` instead :).

schema-admin.graphql
nina-py: I don't know, we could give it a try.

schema-admin.graphql
Herraj: I think what Sri means is that the topics that have `null` as a value, we will map it to `imported` in the data mapping process. So this could stay as a required field

Herraj (dismissed):

schema-admin.graphql
Herraj: We could implement the downloading --> creating a byte stream --> uploading to s3 on the back end/resolver logic.

Downloading from the `http` source is being done on the front end and the resolver for upload image just receives a byte stream. So I think half our work is already done on the backend

jpetto (changes requested):
i'm being a stickler here for the specific backfill use case. i'd rather have tight boundaries here than be flexible for a potential future use.

schema-admin.graphql
```diff
@@ -487,6 +487,92 @@ input CreateApprovedCuratedCorpusItemInput {
...
+    topic: String!
+    """
+    Whether this story is a Pocket Collection.
+    """
+    isCollection: Boolean
+    """
+    A flag to ML to not recommend this item long term after it is added to the corpus.
+    Example: a story covering an election, or "The best of 202x" collection.
+    """
+    isTimeSensitive: Boolean
```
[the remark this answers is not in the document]
Herraj: Yeah if I can recall correctly, we don't have it on the backfill data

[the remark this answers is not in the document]
sri-kirsh: ya we are not sending `isTimeSensitive`for the backfill purpose

schema-admin.graphql
```diff
@@ -487,6 +487,92 @@ input CreateApprovedCuratedCorpusItemInput {
...
+    """
+    createdAt: Int!
+    """
+    A single sign-on user identifier of the user who created this entity.
+    """
+    createdBy: String!
+    """
+    A Unix timestamp of when the entity was last updated.
+    """
+    updatedAt: Int!
```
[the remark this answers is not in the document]
sri-kirsh: we have `updatedAt` as `curated_feed_items.time_updated`. 
we only don't have `updatedBy`

https://getpocket.atlassian.net/wiki/spaces/CP/pages/2642018314/Backfill+Database+Mapping+and+Architecture+Notes#Database-Mapping

schema-admin.graphql
sri-kirsh: ya agree on what herraj mentioned - in the mutation input, we will set the `topic` as `imported` when the topics are null in the legacy database. 

topic will be a required field

schema-admin.graphql
```diff
@@ -487,6 +487,92 @@ input CreateApprovedCuratedCorpusItemInput {
...
+"""
+type ImportApprovedCuratedCorpusItemPayload {
+    """
+    The approved item, as created by an automated process.
+    """
+    approvedItem: ApprovedCuratedCorpusItem!
+    """
+    The scheduled entry, if created by an automated process at the same time.
+    """
+    scheduledItem: ScheduledCuratedCorpusItem
```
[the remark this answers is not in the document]
sri-kirsh: for the backfill purpose, we require the externalId of the scheduledItem. we will use this id to update/delete the scheduledItems during the `datasync` phase

schema-admin.graphql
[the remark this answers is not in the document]
jpetto: ah, okay. for data consistency, i'd suggest sending the `createdBy` value for this field as well. close enough, right? 🙃

schema-admin.graphql
jpetto: what's our plan for this `topic` value? are we going to add it to the acceptable values in the API? will this introduce any issues when viewing/editing these items in the tool?

jpetto: this conversation is happening in slack: https://pocket.slack.com/archives/CTLKW5WCF/p1646934876602779

sri-kirsh: note: this decision would not be blocking this PR as the `topics` are anyway string and would be a mandatory field in the schema in either case.

schema-admin.graphql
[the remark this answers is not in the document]
nina-py: Removing this field 👍. 

@sri-kirsh, @jpetto, I've got a question about `isCollection` and `isSyndicated`. The "Backfill Database Mapping and Architecture Notes" page on Confluence mentions that these don't exist in the legacy database either - are we getting them from Prospect API on the fly during the migration as the Confluence doc seems to indicate or are we removing these from the inputs and setting them to `false`?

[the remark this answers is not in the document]
sri-kirsh: we can have these fields, we can fetch it from prospectApi

schema-admin.graphql
```diff
@@ -521,18 +539,17 @@ input ImportApprovedCuratedCorpusItemInput {
...
-    A topic this story best fits in. The value will be `imported` for migrated items that don't have a topic set.
+    A topic this story best fits in. The value will be `IMPORTED` for migrated items that don't have a topic set.
     """
     topic: String!
     """
-    Whether this story is a Pocket Collection.
+    The source of the corpus item.
     """
-    isCollection: Boolean
+    source: CorpusItemSource!
```
jpetto, line 548: one could argue this mutation implies `BACKFILL`, but i don't have a big problem with this being an input.

schema-admin.graphql
jpetto: i have thrown a wrench into this plan by proposing we instead allow `null` on this database field. ([slack thread](https://pocket.slack.com/archives/CTLKW5WCF/p1647014269434169?thread_ts=1646934876.602779&cid=CTLKW5WCF).) if folks are all on-board/no one sees an issue, we can drop this input.

jpetto (commented):
holding off on approval until we all agree on the topic question, but otherwise this looks good to go to me!

schema-admin.graphql
sri-kirsh: changed it as a nullable field

sri-kirsh (approved):

schema-admin.graphql
sri-kirsh: it might be good to have it explicit, so if data engineering wants to use this mutation later  for non new-tab related loads, they can add their custom enum directly to the source

jpetto (approved):
looks good! thanks for all the conversation everyone - i think we got to the right place 😄

PR #596, opened 2022-03-09
chore(linter): Add infra code to GitHub linter known paths

nina-py:
## Goal

This should get rid of annoying linter warnings on TS files under the `.aws` folder _and_ start to actually run the linter over the code there.

linter-warnings

Apparently negating an ignore pattern is the way to go here. See https://github.com/eslint/eslint/issues/8429#issuecomment-355967308

Bonus: fixed a couple of linting errors in the infra code thanks to the linter no longer ignoring the `.aws` folder.


jpetto (approved):
nice chore fix 😄

PR #601, opened 2022-03-10
feat(corpus): [BACK-1359] Standardize language case

nina-py:
## Goal

Make sure the language code we save for approved and rejected corpus items is always consistent case-wise. Upper case was chosen for language code.

- [x] Added a list of known languages to the config ('EN', 'DE') similar to the work done in Collection API.

- [x] Now checking against this list when creating or updating curated items.

- [ ] Tests outstanding

 
## References

JIRA ticket:
* https://getpocket.atlassian.net/browse/BACK-1359


src/database/helpers/checkLanguage.ts
```diff
@@ -0,0 +1,17 @@
+import { UserInputError } from 'apollo-server';
+import config from '../../config';
+
+/**
+ * Checks if the language code passed in is in the list of allowed languages.
+ *
+ * @param language
+ */
+export const checkLanguage = (language: string): void => {
+  if (!config.app.languages.includes(language)) {
```
Herraj, line 10: I think `.includes()` is case sensitive but I'm not sure. If it is then see the comment below

src/database/mutations/ApprovedItem.ts
```diff
@@ -21,6 +21,9 @@ export async function createApprovedItem(
   return db.approvedItem.create({
     data: {
       ...data,
+      // Make sure the language code is in upper case.
+      language: data.language?.toUpperCase(),
```
Herraj, line 25: Wondering why we need to make it to upper case here if the helper function checks in the resolver that we only receive the correct language code i.e uppercase EN/DE

Herraj (commented):
Couple of comments for clarification. Otherwise looks good! I like that you've put it in the config variable

src/database/helpers/checkLanguage.ts
nina-py: Good point @Herraj, I was making the updates in different areas of the code without thinking what's going to be executed in which order 🙃. I'll remove those `'toUpperCase()` calls as they should never be needed.

.circleci/config.yml
```diff
@@ -184,6 +184,7 @@ jobs:
         auth:
           username: $DOCKERHUB_USERNAME
           password: $DOCKERHUB_PASSWORD
+    parallelism: 4
```
Herraj, line 187: 🤔

Herraj (approved):

.circleci/config.yml
jpetto: yeah 🤔. will this cause issues with integration tests expecting strict database counts/values? or does this mean a whole separate mysql instance is spun up to parallelize the tests? (the [circle docs](https://circleci.com/docs/2.0/parallelism-faster-jobs/) aren't helping me much here 🙃 )

src/admin/resolvers/mutations/ApprovedItem.ts
```diff
@@ -180,7 +187,8 @@ export async function rejectApprovedItem(
     url: approvedItem.url,
     title: approvedItem.title,
     topic: approvedItem.topic,
-    language: approvedItem.language,
+    // Make sure the language code is in upper case.
+    language: approvedItem.language.toUpperCase(),
```
jpetto, line 191: if this data is coming out of the db (and we're enforcing case going _in_ to the db), should we ever need to uppercase this value?

jpetto (approved):
i think we can merge this as-is, but had a couple non-necessarily-blocking questions.

src/admin/resolvers/mutations/ApprovedItem.ts
Herraj: Yeah that's what my previous comment also mentions. If we never allow to a non lowercase value be created in the first place, we should only have uppercase values

.circleci/config.yml
nina-py: Oh, I forgot to amend the commit to mention this. It's from a tip on Circle CI: https://app.circleci.com/pipelines/github/Pocket/curated-corpus-api/2672/workflows/a46f14e1-f21d-430c-8745-79e31174879c/jobs/11765. I thought I'd give it a try and see if anything breaks. It doesn't 🙃. However, it doesn't seem to speed the tests up at all - they still run for ~2 mins, but now in four different containers. I'll merge this one as-is but make a note to take this out in a follow-up PR.

nina-py: Update: nope, I'll do it here since I need to update another part of code anyway.

src/admin/resolvers/mutations/ApprovedItem.ts
nina-py: Oh, that is my mistake. I thought I took out all of these out in a previous commit.

Herraj (approved):

PR #609, opened 2022-03-13
feat(schema): [INFRA-374] Add `source` field to approved item schema

nina-py:
## Goal

Add a new field to distinguish approved items backfilled from the legacy database from everything else that will be added through the frontend by the curators in the future.

- Added the `source` field to Prisma schema and generated a database migration with the default value of `PROSPECT` for the newly created field.

- Added the `source` field to Snowplow events as an optional value to be sent through with relevant events.

- Updated the README to add that before running a migration you need to start the app in Docker - this catches me every time.
- **Note**: waiting on Snowplow schema updates for all tests to pass.

JIRA ticket:
* https://getpocket.atlassian.net/browse/INFRA-374


prisma/schema.prisma
```diff
@@ -24,24 +24,25 @@ enum CuratedItemSource {
...
+  prospectId      String?          @db.VarChar(255)
+  url             String           @db.VarChar(500)
+  title           String           @db.VarChar(255)
+  excerpt         String           @db.Text
+  status          CuratedStatus    @default(CORPUS)
+  language        String           @db.VarChar(2)
+  publisher       String           @db.VarChar(255)
+  imageUrl        String           @db.VarChar(500)
+  topic           String           @db.VarChar(255)
+  source          CorpusItemSource @default(PROSPECT)
```
jpetto, line 38: i'm leaning on not having a default here, as `MANUAL` will be a common use case in the tool. i.e. i think the mutations will have to receive/specify this value.

thoughts?

jpetto (changes requested):
a few questions around required/default data, but this is really close 😄

prisma/schema.prisma
jpetto: hmm, is this an existing data problem? i think we could work around it by customizing the migration:

1. add the field to the table as not required
2. write some SQL to update all rows with a default value of `PROSPECT`
3. update the field to required

i know it's nit-picky, but i don't think we should let prisma dictate our db design 😄

src/events/snowplow/ReviewedItemSnowplowHandler.ts
```diff
@@ -96,6 +97,11 @@ export class ReviewedItemSnowplowHandler extends CuratedCorpusSnowplowHandler {
       corpusReviewStatus = CorpusReviewStatus.REJECTED;
     }
 
+    let itemSource: CorpusItemSource | undefined;
```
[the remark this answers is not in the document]
jpetto: ah, good point!

this does make me wonder if we need a `source` column for rejected items as well, but we can cross that bridge when we come to it.

prisma/schema.prisma
nina-py: Hi @jpetto, existing data doesn't really hold us back (since there is none in production yet), however Prisma's way of doing things is. I've done two commits today - one with the Snowplow events fixed up and the migration as I set it up initially, with the default source as `PROSPECT`, and the next one with two migrations that initially set up the `source` field as not required, update any existing entries and then make the field required without a default value.

What this does is screw everything else up - specifically, the mutations that now require the value for the `source` field to be magicked from somewhere. I've tried and subsequently binned various approaches to fixing this but it's not so simple. Maybe I'm not seeing something very obvious but from what I've learned today is that the most practicaly way to keep Prisma's generated types play nicely with the rest of the code is to pass the `source` value as one of mutation input values. This in turn breaks the frontend straight away. 

The new `import` mutation accepts the source value, but the original `createApprovedItem` doesn't. So a decision needs to be made if this mutation, and subsequently the frontend, need to be updated if we want to keep the `source` value required but without a default value set at the DB level.

jpetto:
unfortunately, it looks like two tickets (INFRA-331 and INFRA-374) ended up overlapping. @kkyeboah has #615 ready to go, which i think completes all the work in this PR 😕.

nina-py:
Oh no. I'm a very sad panda right now.

PR #615, opened 2022-03-15
feat: implement 'importApprovedCuratedCorpusItem' mutation

kkyeboah:
## Goal
Implement the `importApprovedCuratedCorpusItem` mutation resolver.

- Is `PROSPECT` the correct default value for `source`?
- Anything I may have missed

## Implementation Decisions
- It was a bit of a challenge following the established patterns in the codebase which works under the premise that the GraphQL types will map directly to the database types. 

## Deployment steps
- [ ] Coordinate with analytics to get `iglu:com.pocket/reviewed_corpus_item/jsonschema/1-0-2` schema to production

## References

JIRA ticket: INFRA-331


prisma/schema.prisma
```diff
@@ -24,24 +24,25 @@ enum CuratedItemSource {
...
+  prospectId      String?            @db.VarChar(255)
+  url             String             @db.VarChar(500)
+  title           String             @db.VarChar(255)
+  excerpt         String             @db.Text
+  status          CuratedStatus      @default(CORPUS)
+  language        String             @db.VarChar(2)
+  publisher       String             @db.VarChar(255)
+  imageUrl        String             @db.VarChar(500)
+  topic           String             @db.VarChar(255)
+  source          CuratedItemSource  @default(PROSPECT)
```
kkyeboah, line 38: Is this the correct default?

src/admin/resolvers/mutations/ApprovedItem.ts
```diff
@@ -245,3 +258,105 @@ export async function uploadApprovedItemImage(
...
+  context: IContext
+): Promise<ImportApprovedCuratedCorpusItemPayload> {
+  // check if user is authorized to import an item
+  if (!context.authenticatedUser.canWriteToCorpus()) {
+    throw new AuthenticationError(ACCESS_DENIED_ERROR);
+  }
+
+  let approvedItem = await getApprovedItemByUrl(context.db, data.url);
+  if (!approvedItem) {
+    approvedItem = await dbImportApprovedItem(
```
sri-kirsh, line 283: we should also be calling the `checkLanguage()` function that's called in ApprovedItem mutation: https://github.com/Pocket/curated-corpus-api/blob/main/src/admin/resolvers/mutations/ApprovedItem.ts#L75

kkyeboah: so we should be throwing an error if the language is not valid, correct?

src/admin/resolvers/mutations/ApprovedItem.integration.ts
```diff
@@ -860,6 +863,172 @@ describe('mutations: ApprovedItem', () => {
...
   });
+
+  describe('importApprovedCuratedCorpusItem mutation', () => {
+    const input: ImportApprovedCuratedCorpusItemInput = {
+      url: 'https://test.com/docker',
+      title: 'Find Out How I Cured My Docker In 2 Days',
+      excerpt: 'A short summary of what this story is about',
+      status: CuratedStatus.CORPUS,
+      imageUrl: 'https://test.com/image.png',
+      language: 'EN',
```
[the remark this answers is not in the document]
kkyeboah: do you mean to update the schema to accept an enum for `EN` or `DE` or update the test input data to `en`?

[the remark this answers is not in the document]
sri-kirsh: no just updating the field description that `this field accepts only EN or DE as language input`

src/admin/resolvers/mutations/ApprovedItem.ts
kkyeboah: done

src/admin/resolvers/mutations/ApprovedItem.integration.ts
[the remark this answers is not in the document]
kkyeboah: done

src/admin/resolvers/mutations/ApprovedItem.integration.ts
```diff
@@ -860,6 +863,172 @@ describe('mutations: ApprovedItem', () => {
...
+      eventEmitter.on(
+        ScheduledCorpusItemEventType.ADD_SCHEDULE,
+        addScheduleEventTracker
+      );
+
+      const headers = {
+        groups: `${MozillaAccessGroup.SCHEDULED_SURFACE_CURATOR_FULL}`,
+      };
+      const server = getServerWithMockedHeaders(headers, eventEmitter);
+      await server.start();
```
jpetto, line 939: i don't think we need `server.start()` if we're only using `executeOperation`.

src/admin/resolvers/mutations/ApprovedItem.ts
```diff
@@ -245,3 +258,108 @@ export async function uploadApprovedItemImage(
...
+): Promise<ImportApprovedCuratedCorpusItemPayload> {
+  // check if user is authorized to import an item
+  if (!context.authenticatedUser.canWriteToCorpus()) {
+    throw new AuthenticationError(ACCESS_DENIED_ERROR);
+  }
+
+  // Check the language. Throw an error if not valid
+  checkLanguage(data.language);
+
+  let approvedItem = await getApprovedItemByUrl(context.db, data.url);
```
jpetto, line 284: i don't know how much it matters, but what do you think about inverting this logic? instead of checking existence up front (which might have a high performance cost?), we instead catch the already exists error and retrieve the existing id in the `catch` block? (same for scheduled item.)

sri-kirsh (dismissed):

src/database/mutations/ScheduledItem.ts
```diff
@@ -44,6 +45,23 @@ export async function createScheduledItem(
...
+ * @param db
+ * @param data
+ */
+export async function importScheduledItem(
+  db: PrismaClient,
+  data: ImportScheduledItemInput
+): Promise<ScheduledItem> {
+  return db.scheduledItem.create({
+    data,
+    include: {
```
jpetto, line 59: as we're returning the approved item as well, do we need this `include`?

src/database/queries/ScheduledItem.ts
```diff
@@ -117,3 +117,26 @@ export async function getItemsForScheduledSurface(
...
+): Promise<ScheduledItem | null> {
+  return db.scheduledItem.findUnique({
+    where: {
+      ItemScheduledSurfaceDate: {
+        approvedItemId: data.approvedItemId,
+        scheduledSurfaceGuid: data.scheduledSurfaceGuid,
+        scheduledDate: data.scheduledDate,
+      },
+    },
+    include: {
```
jpetto, line 138: same here - do we need this `include`?

jpetto (changes requested):
a migration request that is nit-picky, and then a few potential performance questions. 😄

prisma/schema.prisma
sri-kirsh: `prospect` as default value for `source` would work in automatedmutation case as we are explicitly setting this value

but if a curator adds an item manually, how will the `createApprovedItem` mutation will set this to `manual`?

src/database/mutations/ScheduledItem.ts
kkyeboah: I was battling with types and took the path of least resistance, I will revisit

src/database/queries/ScheduledItem.ts
kkyeboah: same comment as above

src/admin/resolvers/mutations/ApprovedItem.ts
kkyeboah: I think that makes sense - This is a new service so very high chance that all the records will not exist. In that case, attempting to create first, definitely saves us some database calls. 

But that also holds true for the opposite - what happens when we already have most of the records but need to rerun this again?

prisma/schema.prisma
jpetto: this new field will require changes in the curation tool. i'll make a ticket for that now.

src/admin/resolvers/mutations/ApprovedItem.ts
jpetto: i _think_ if we have to re-run a backfill, the assumption is we'd delete everything for that new tab first.

src/admin/resolvers/mutations/ApprovedItem.integration.ts
```diff
@@ -860,6 +863,172 @@ describe('mutations: ApprovedItem', () => {
       );
     });
   });
+
+  describe('importApprovedCuratedCorpusItem mutation', () => {
+    const input: ImportApprovedCuratedCorpusItemInput = {
+      url: 'https://test.com/docker',
+      title: 'Find Out How I Cured My Docker In 2 Days',
+      excerpt: 'A short summary of what this story is about',
+      status: CuratedStatus.CORPUS,
```
sri-kirsh, line 872: not a blocker: just calling out that we will be setting `RECOMMENDATION` as default for all our backfill mutations

prisma/schema.prisma
sri-kirsh: - we can make the `source` as nullable field without default to unblock the PR
- we will have follow up ticket (Created by jonathan) to add `source` to other mutations
- then we set a default to source (Depending on product team needs) - but this can happen outside this work

src/database/mutations/ScheduledItem.ts
kkyeboah: We need this one for the snowplow event. We could rework that but that would increase the scope of this work by a lot

src/database/queries/ScheduledItem.ts
kkyeboah: need this to satisfy the type GraphQL return type. Happy to follow up and simplify in a separate PR

prisma/schema.prisma
kkyeboah: implemented per @sri-kirsh's comment above

jpetto (approved):
lgtm 👍

src/admin/resolvers/mutations/ApprovedItem.ts
```diff
@@ -245,3 +259,123 @@ export async function uploadApprovedItemImage(
...
+    topic: data.topic,
+    url: data.url,
+    isCollection: data.isCollection,
+    isSyndicated: data.isSyndicated,
+    source: data.source,
+    createdAt: fromUnixTime(data.createdAt),
+    createdBy: data.createdBy,
+    updatedAt: fromUnixTime(data.updatedAt),
+    updatedBy: data.updatedBy,
+  };
```
Herraj, line 361: nit: if we wanted to make this short I think we could've done this but this is fine too
`return { ...data, createdAt: ..., updatedAt: ...}`

Herraj (approved):
:shipit:

PR #621, opened 2022-03-16
chore: use GraphQL Enum for language validation

kschelonka:
This ensures language constraints are documented
in the schema, and validation is automatic.

This was marked as an unsafe field change in apollo for some specific queries, but it should be fine... please confirm.
https://studio.apollographql.com/graph/pocket-admin-api/operationsCheck/a3ff3fe6-143f-4b2c-96a9-fe23834b14cb?variant=current


kschelonka:
If you have constraints that are not visible at the schema level, it makes it much more difficult for clients. You have to also remember to maintain both the schema documentation as well as whatever configuration/checks inside the code.

kschelonka:
Since I don't have project context, I'm not understanding what the admin API issues would be. Can you please elaborate?

jpetto:
been consider this and here's where i landed:

- let's do language validation at the schema level as is done in this PR. as long as we have these enums "namespaced", our flexibility is maintained and we will avoid graph collisions.
- i think having both input and output use the enum is correct, as we shouldn't ever send values that aren't consistent with what we receive.
- let's rename the enum to `CorpusLanguage` (there have been conversations elsewhere that happened well after this repo was up and running that "Curated Corpus" is overly descriptive - this will be _the_ corpus).
- we'll make a ticket to update collections to do this same schema validation.

is that cool with everyone?

kschelonka:
How about I just keep this PR out here for someone to iterate on as needed once requirements have been decided?

nina-py:
Thanks @kat, @jpetto - I can take this PR over and see it through to completion.

nina-py:
The warnings in the Apollo checks should be fine. The changes from `main` have been merged in and linting errors are fixed.

jpetto (approved):
thank you @kat and @nina-py for putting this together 😄

kkyeboah (approved):

PR #627, opened 2022-03-18
fix: upload image to S3 as part of import mutation

kkyeboah:
## Goal

Upload image to S3 before creating an approved item.

## References

JIRA ticket: INFRA-396

### DIFF
--- a/src/admin/resolvers/mutations/ApprovedItem.ts
+++ b/src/admin/resolvers/mutations/ApprovedItem.ts
@@ -13,7 +13,7 @@ import {
   ReviewedCorpusItemEventType,
   ScheduledCorpusItemEventType,
 } from '../../../events/types';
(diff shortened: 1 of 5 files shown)


src/admin/resolvers/mutations/ApprovedItem.ts
```diff
@@ -290,6 +295,11 @@ export async function importApprovedItem(
       context.db,
       data.url
     )) as ApprovedItem;
+
+    // If there's an invalid image, stop here and throw an exception
+    if (!approvedItem && e instanceof InvalidImageUrl) {
```
sri-kirsh, line 300: nice! if we have an existing approvedItem, we can just return that - otherwise, we are trying to add a new approvedItem with invalid image, so we throw userInputError

sri-kirsh (approved):

PR #644, opened 2022-03-22
feat(scheduledItem): add scheduledSurfaceGuid as a field to ScheduledCuratedCorpusItem

sri-kirsh:
## Goal

- add `scheduledSurfaceGuid` as a required field to the entity `ScheduledCuratedCorpusItem`

Ticket: https://getpocket.atlassian.net/browse/INFRA-337


schema-admin.graphql
```diff
@@ -345,6 +345,10 @@ type ScheduledCuratedCorpusItem {
     The associated Approved Item.
     """
     approvedItem: ApprovedCuratedCorpusItem!
+    """
+    Optional, The GUID of this schedudedSurface to which this item is scheduled. Example: 'NEW_TAB_EN_US'.
```
jpetto, line 349: we should be able to make this required - it's required in the db schema.

schema-admin.graphql
```diff
@@ -345,6 +345,10 @@ type ScheduledCuratedCorpusItem {
...
+    """
+    scheduledSurfaceGuid: ID
```
nina-py, line 351: We'll never have a scheduled item without a surface it was scheduled on, so I think it's safe to make this prop required

src/admin/resolvers/mutations/ApprovedItem.integration.ts
```diff
@@ -8,24 +8,24 @@ import {
   createApprovedItemHelper,
   createRejectedCuratedCorpusItemHelper,
   createScheduledItemHelper,
-  getServerWithMockedHeaders,
+  getServerWithMockedHeaders
```
sri-kirsh, line 11: lot of prettier changes in this file

src/admin/resolvers/mutations/ApprovedItem.integration.ts
```diff
@@ -973,6 +973,10 @@ describe('mutations: ApprovedItem', () => {
       );
       expect(approvedItem.url).to.equal(input.url);
       expect(approvedItem.imageUrl).to.match(urlPattern);
+      expect(approvedItem.externalId).is.not.undefined;
```
sri-kirsh, line 976: this is the actual change, added assertions for fields we expect in the backfill code

nina-py (approved):

PR #649, opened 2022-03-25
chore(scheduledItem): make scheduledSurfaceGuid as an index to scheduledItem table

sri-kirsh:
### Goal
make scheduledSurfaceGuid as an index to scheduledItem table

this will allow us to search rows faster in the destination database when we migrate them.


jpetto (approved):
:shipit:

kkyeboah (approved):

PR #651, opened 2022-03-28
feat: send add-scheduled-item event to event bus

kschelonka:
## Goal

Build add-scheduled-item event and send to event bus.

Note that no special permissions are required to send to the default event bus in an account.

The code is structured as follows:

* Function `eventBusHandler` to initialize the handler in `main.ts`, following existing pattern
* Configuration in `EventBusConfig.ts` which defines callback(s) for generating payload, and a mapping of event name to payload generator function. This is used by default if a configuration is not explicitly passed
* The EventBusHandler class itself, which is just a generic class that:
    1. contains the logic for sending an event payload to eventbridge event bus (based on app configuration)
    2. registers all listeners on the mapping of `eventName: callback` passed to the constructor (or the default mapping in `EventBusConfig.ts`), and invokes the function to send to event bus

- any 

## Implementation Decisions
- separate the payload generator functions from the event handler class for clarity

## References

JIRA ticket:
* [INFRA-355]

[INFRA-355]: https://getpocket.atlassian.net/browse/INFRA-355?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.aws/src/config/index.ts
```diff
@@ -37,6 +37,10 @@ export const config = {
     repository: 'pocket/curated-corpus-api',
     branch,
   },
+  // The name of the Event Bus to publish events to (per account)
```
sri-kirsh, line 40: we need  an `iam` to publish from `curatedCorpusApi` to the event bus. 
https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-event-bus-perms.html#eb-event-bus-example-policy-cross-account

also can we test this in dev to make sure the integration from curatedCorpusApi -> eventBus works

kschelonka: as far as I can tell, no special permission is required to publish to the default event bus for a given AWS account

src/events/eventBus/EventBusHandler.ts
```diff
@@ -0,0 +1,108 @@
...
+   * @param eventPayload the payload to send to event bus
+   */
+  async sendEvent(eventPayload: BaseEventBusPayload) {
+    const putEventCommand = new PutEventsCommand({
+      Entries: [
+        {
+          EventBusName: config.aws.eventBus.name,
+          Detail: JSON.stringify(eventPayload),
+          Source: config.eventBridge.source,
+          DetailType: config.eventBridge.addScheduledItemEventType,
```
kkyeboah, line 90: This will always send the same `detail-type`. We need to send different `detail-type`s based on the action.
Ex. `update-scheduled-item` and `delete-scheduled-item`

src/shared/utils.ts
```diff
@@ -29,3 +29,13 @@ export function groupBy(array: any[], key: string) {
 
   return result;
 }
+
+/**
+ * Converts a Date object to a YYYY-MM-DD string (in UTC)
+ */
+export function toUtcDateString(date: Date) {
```
kkyeboah, line 36: not opposed to this because it's so small but we could throw a library in here and not maintain it `date-fns` is a good one if this repo does not already have one. 

Also, this is not a blocking comment.

src/events/eventBus/EventBusHandler.ts
kschelonka: good catch, I forgot to update that

kschelonka: it's done

.aws/src/config/index.ts
kschelonka: I was wrong, tested with updated permissions

kkyeboah (approved):

PR #652, opened 2022-03-29
feat(events): Send remove-schedule-item event to event bridge

kkyeboah:
## Goal

Implement remove-schedule-item event

- 

## Implementation Decisions
- Reusing the payload transformer since the event body is always the same

## References

JIRA ticket:
* [INFRA-357]

Documentation:
* [Project doc](https://getpocket.atlassian.net/wiki/spaces/CP/pages/2648178701/Curation+Migration+DataSync+Events)

[INFRA-357]: https://getpocket.atlassian.net/browse/INFRA-357?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/config/index.ts
+++ b/src/config/index.ts
@@ -47,6 +47,7 @@ export default {
   },
   eventBridge: {
     addScheduledItemEventType: 'add-scheduled-item',
+    removeScheduledItemEventType: 'remove-scheduled-item',
     source: 'curation-migration-datasync',
   },
   sentry: {
--- a/src/events/eventBus/EventBusConfig.ts
+++ b/src/events/eventBus/EventBusConfig.ts
@@ -11,8 +11,18 @@ import { toUtcDateString } from '../../shared/utils';
 // Used to configure the default EventBusHandler class,
 // if a mapping is not passed to the constructor.
 export const eventBusConfig: EventHandlerCallbackMap = {
-  [ScheduledCorpusItemEventType.ADD_SCHEDULE]: (data: any) =>
-    payloadBuilders.scheduledItemEvent(data),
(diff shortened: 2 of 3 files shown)


src/config/index.ts
```diff
@@ -47,6 +47,7 @@ export default {
   },
   eventBridge: {
     addScheduledItemEventType: 'add-scheduled-item',
+    removeScheduledItemEventType: 'remove-scheduled-item',
```
sri-kirsh, line 50: the `detail-type` defined in event bridge rule is `delete-scheduled-item`

src/events/eventBus/EventBusConfig.ts
```diff
@@ -22,14 +32,16 @@ const payloadBuilders = {
...
    * @param data ScheduledCorpusItemPayload
    * @returns ScheduledItemEventBusPayload
    */
   scheduledItemEvent(
+    eventType: string,
     data: ScheduledCorpusItemPayload
   ): ScheduledItemEventBusPayload {
-    const eventPayload: ScheduledItemEventBusPayload = {
-      eventType: config.eventBridge.addScheduledItemEventType,
+    return {
```
sri-kirsh, line 43: just note:  we are re-using the entire body of the `add`, but we will be really consuming only scheduled item externalIds.

when a scheduled item gets deleted, do we also get other details like title, etc - are they all mandatory fields in the payload?

kkyeboah: we do from the mutation

sri-kirsh (approved):

PR #653, opened 2022-03-29
fix(shared data): [BACK-1381] Cannot add approved item with `CORONAVIRUS` topic

nina-py:
## Goal

Let the curators save/edit items with `CORONAVIRUS` as the topic. While present in the GraphQL schema, for some reason it was overlooked when a TypeScript enum was set up with the same data.

## References

JIRA ticket:
* https://getpocket.atlassian.net/browse/BACK-1381


jpetto (approved):
yay for simple bug fixes 😄

PR #659, opened 2022-03-29
feat(events): update events

kschelonka:
## Goal

Send update events for scheduled + approved items.

Since the application emits two different kinds of events for updating (updating the scheduled item's date/surface vs. the underlying item's fields), we handle them as separate events for event bridge.

Since the legacy system does not have a concept of an approved item that isn't scheduled, It is up to the event subscriber to determine whether it needs to process changes to an approved item.

- all

## Implementation Decisions
- Not refactoring the types, although at this time only approved items (and not a union of approved + rejected) are included as a payload on the update event

## References

JIRA ticket:
* [INFRA-356]

Documentation:
* [Project doc](https://getpocket.atlassian.net/wiki/spaces/CP/pages/2648178701/Curation+Migration+DataSync+Events)

[INFRA-356]: https://getpocket.atlassian.net/browse/INFRA-356?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/events/eventBus/EventBusConfig.ts
+++ b/src/events/eventBus/EventBusConfig.ts
@@ -1,4 +1,7 @@
 import {
+  ApprovedItemEventBusPayload,
+  ReviewedCorpusItemEventType,
+  ReviewedCorpusItemPayload,
   ScheduledCorpusItemEventType,
   ScheduledCorpusItemPayload,
   ScheduledItemEventBusPayload,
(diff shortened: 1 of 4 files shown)


nina-py (approved):
Looks good!

kkyeboah (approved):

src/events/eventBus/EventBusConfig.ts
```diff
@@ -59,4 +74,30 @@ const payloadBuilders = {
...
+    data: ReviewedCorpusItemPayload
+  ): ApprovedItemEventBusPayload {
+    // The nullish coalesce and checking for properties are due to
+    // union type in ReviewedCorpusItemPayload
+    const item = data.reviewedCorpusItem;
+    return {
+      eventType: eventType,
+      approvedItemId: item.externalId,
+      url: item.url,
+      title: item.title ?? undefined,
```
sri-kirsh, line 88: we have to check them for undefined coz the rejectedItems won't have these field set?

sri-kirsh (approved):

src/events/eventBus/EventBusConfig.ts
kschelonka: yep

PR #669, opened 2022-04-01
refactor(types): remove extra lookup types

jpetto:
## Goal

remove extra lookup types. this refactor removes three extra types that mapped scheduled surfaces to mozillians access groups. imo, these extra types were all interdependent and difficult to reason about.

- associate mozillians access groups directly with scheduled surfaces
- move some stuff out of types into utils

- adding `accessGroup` to the shared `ScheduledSurface` type. do we like it?


Herraj (approved):
:shipit:

PR #672, opened 2022-04-06
feat(scheduledItem): sort scheduled items by createdAt asc

jpetto:
## Goal

sort scheduled items by `updatedAt 'asc'`.

- sorting for both admin and public queries

- do we need both public and admin queries sorted?

## References

JIRA ticket:
* https://getpocket.atlassian.net/browse/BACK-1391

### DIFF
--- a/src/database/queries/ScheduledItem.ts
+++ b/src/database/queries/ScheduledItem.ts
@@ -28,7 +28,9 @@ export async function getScheduledItems(
 
   // Get a flat array of scheduled items from Prisma
   const items = await db.scheduledItem.findMany({
-    orderBy: { scheduledDate: 'asc' },
+    // we need to order by scheduleDate first, as we perform a programmatic
+    // groupBy below on that field
+    orderBy: [{ scheduledDate: 'asc' }, { updatedAt: 'asc' }],
     where: {
       scheduledSurfaceGuid: { equals: scheduledSurfaceGuid },
       scheduledDate: {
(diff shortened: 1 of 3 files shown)


sri-kirsh:
does public here mean anyone with `sso` but not assigned to a cognito group to schedule an item under any surfaceGuid?

kschelonka (approved):
lgtm from code perspective -- can wait to confirm on acceptance criteria

jpetto:
> does `public` here mean anyone with `sso` but not assigned to a cognito group to schedule an item under any surfaceGuid?

good question! `public` here means the query exposed via client api, so just how results are returned to clients. no mutation capability!

src/database/queries/ScheduledItem.ts
```diff
@@ -28,7 +28,9 @@ export async function getScheduledItems(
 
   // Get a flat array of scheduled items from Prisma
   const items = await db.scheduledItem.findMany({
-    orderBy: { scheduledDate: 'asc' },
+    // we need to order by scheduleDate first, as we perform a programmatic
+    // groupBy below on that field
+    orderBy: [{ scheduledDate: 'asc' }, { createdAt: 'asc' }],
```
nina-py, line 33: I was wondering if we could use `updatedAt` instead to circumvent the need to add a `manualSort` field.

nina-py (approved):

PR #675, opened 2022-04-12
fix: rename event bus payload property

kkyeboah:
## Goal

Rename event properties to match expectations in data sync service:
`scheduledItemId` => `scheduledItemExternalId`
`approvedItemId` => `approvedItemExternalId`

## References

JIRA ticket: INFRA-393


nina-py (commented):
👍 to keeping all the prop names in sync.

PR #684, opened 2022-05-09
fix(query): [BACK-1433] Public query returns too many results

nina-py:
## Goal

Investigate the public query that returned too many results (or, for Pocket Hits, results _at all_). 

- It turns out the `items` subquery within the `scheduledSurface` query never got the correct value for scheduled surface GUID. Now fixed.

- Added an integration tests that checks that the code is really fixed.

## References

JIRA ticket:
* https://getpocket.atlassian.net/browse/BACK-1433


Herraj (approved):

jpetto (approved):
short and sweet!

PR #685, opened 2022-05-09
feat(authors): add authors data to schemas

jpetto:
## DO NOT MERGE
Until the corresponding changes to the frontend in https://getpocket.atlassian.net/browse/BACK-1427 are ready to go.

## Goal

add authors data to schemas - public and admin API, prisma, and snowplow. 
+ update all the resolvers to take in/return authors.

RESOLVED - **note** we are waiting on a confirmation on if we need to send `sortOrder` to snowplow. there's a TODO in that file in this PR ([slack thread](https://pocket.slack.com/archives/CTLKW5WCF/p1652133701280769?thread_ts=1651704195.064879&cid=CTLKW5WCF)).

- this is modeled off our work in collections. anything we could do better?

JIRA ticket:
* https://getpocket.atlassian.net/browse/BACK-1423
* https://getpocket.atlassian.net/browse/BACK-1426


nina-py:
@jpetto, I've added a migration for changes in Prisma schema and, as I was going about trying to fix the tests I veered into updating the resolvers - another ticket, as it turns out! I'm almost there, but a few tests are still erroring out.

src/database/mutations/ScheduledItem.ts
```diff
@@ -40,7 +40,13 @@ export async function createScheduledItem(
...
     },
     include: {
-      approvedItem: true,
+      approvedItem: {
+        include: {
+          authors: {
+            orderBy: [{ sortOrder: 'asc' }],
+          },
+        },
+      },
```
nina-py, line 49: Despite _seemingly_ pulling in `authors` from Prisma, integration tests for all `ScheduledItem`-related mutations are failing to find the `authors` prop on the included `approvedItem`.

src/admin/resolvers/mutations/ApprovedItem.ts
```diff
@@ -144,6 +145,18 @@ export async function updateApprovedItem(
     );
   }
 
+  // TODO: cleanly delete existintg authors as we do in collections
+  // get collectionStory internal id for deleting authors & checking for existing story
+  const existingItem = await getApprovedItemByUrl(context.db, data.url);
```
jpetto, line 150: this could be a new function `getApprovedItemByExternalId`? or do we think getting by URL (which is unique!) is fine?

nina-py: I opted for retrieving this via `getApprovedItemByExternalId`, the one that retrieves by URL is used to query when an item is added manually. I personally feel more comfortable using the id value rather than URL :).

schema-admin.graphql
```diff
@@ -151,6 +151,10 @@ type ApprovedCorpusItem @key(fields: "url") {
     """
     excerpt: String!
     """
+    The authors associated with this ApprovedCorpusItem.
+    """
+    authors: [CorpusItemAuthor!]
```
kschelonka, line 156:
```suggestion
    authors: [CorpusItemAuthor!]!
```

?

src/admin/resolvers/mutations/ScheduledItem.integration.ts
```diff
@@ -204,6 +210,22 @@ describe('mutations: ScheduledItem', () => {
...
+      // a custom comparison
+      if (approvedItemAuthors) {
+        const approvedItemAuthorsMapped = approvedItemAuthors.map((aia) => {
+          return {
+            name: aia.name,
+            sortOrder: aia.sortOrder,
+          };
+        });
+
+        expect(approvedItemAuthorsMapped).to.deep.equal(authorsReturned);
```
kschelonka, line 226: jest has this built in: https://jestjs.io/docs/expect#expectobjectcontainingobject

src/admin/resolvers/mutations/ApprovedItem.ts
```diff
@@ -144,6 +145,25 @@ export async function updateApprovedItem(
     );
   }
 
+  // To be able to delete authors associated with a corpus item, we first need
+  // to get the internal (integer) id for the story. This means doing a DB query
+  // to fetch the entire object.
+  const existingItem = await getApprovedItemByExternalId(
```
kschelonka, line 151: Doesn't the association table have `externalId`?

prisma/schema.prisma
```diff
@@ -91,8 +102,8 @@ model ScheduledItem {
   approvedItem ApprovedItem @relation(fields: [approvedItemId], references: [id])
 
   // indexes
-  @@index ([scheduledSurfaceGuid])
   // Prevent scheduling the same item on the same Scheduled Surface for the same day multiple times.
   @@unique([externalId])
   @@unique([approvedItemId, scheduledSurfaceGuid, scheduledDate], name: "ItemScheduledSurfaceDate")
+  @@index([scheduledSurfaceGuid])
```
Herraj, line 108: nit pick: I think this go accidentally moved from line 105

Can be fixed later on!

src/admin/resolvers/mutations/ApprovedItem.ts
kschelonka: oh nvm... I think from reading the schema, that's another "external" ID for that specific author-item relationship?

I don't know prisma, but I feel like you *must* be able to delete by association... or add the additional FK

src/database/queries/ApprovedItem.ts
```diff
@@ -32,12 +32,17 @@ export async function getApprovedItems(
   const baseArgs: prisma.Prisma.ApprovedItemFindManyArgs = {
     orderBy: { createdAt: 'desc' },
     where: constructWhereClauseFromFilters(filters),
+    include: {
+      authors: {
+        orderBy: [{ sortOrder: 'asc' }],
```
Herraj, line 37: Do we need to provide the `sortOrder` for **finding** and **deleting** items from the DB? The fact that we save them to the DB with the `asc` `sortOrder`, shouldn't the item be returned in the same sort order?

Unless it is a prisma thing where do find and delete, it needs to match the parameters of how it was saved?

Herraj (dismissed):
I'm tired so couldn't give you a fully focused review -- if you want to wait until tomorrow, I can review this first thing tomorrow morning and merge it.

But anyways, looks good to me 😄

kschelonka (approved):
seems like you'd get an easy win adding another FK relationship for approvedItem.externalId to authors, but I see no reason to not ship

src/config/index.ts
```diff
@@ -42,7 +42,6 @@ export default {
   },
   events: {
     source: 'curated-corpus-api',
-    // TODO: what should this value be? See list-api, user-api with similar comments
     version: '0.0.2',
```
bassrock, line 45: Long term we should see if analytics will let us make this the GIT SHA

src/admin/resolvers/mutations/ApprovedItem.ts
nina-py: The way this API is built (and also `collection-api`), with extra "external" ids to avoid exposing internal integer ones, means we can't actually delete related records by association until we get the full approved item record from the database that contains the integer primary key. 

It's the admin/internal part of the API so the performance hit of fetching this record an extra time is negligible. I would do it differently if performance was important.

src/database/queries/ApprovedItem.ts
nina-py: @Herraj, we definitely need to provide `sortOrder` when fetching a full `ApprovedItem` record. Even if records are written to a relational database sequentially, there is no guarantee they will be returned in that order unless you specify it explicitly. 

I am not sure about providing `sortOrder` when _deleting_ items. That shouldn't happen anywhere in the codebase.  The closest to this that I can think of is in the `deleteApprovedItem` mutation where we fetch the entire record, including the authors sorted correctly, _before_ we go ahead and hard delete everything - this is done to return the now deleted approved item as the mutation result for consistency with other mutations.

Herraj (approved):

PR #687, opened 2022-05-17
feat(history): [BACK-1439] Add `scheduledSurfaceHistory` subquery to ApprovedItem

nina-py:
## Goal

We need to add a query that retrieves scheduled entries for a corpus item. 

- Added a subquery to the admin schema.

- Added a resolver that takes in result limit + scheduled surface guid, both optional, and retrieves an array of scheduled items in return.

- [x] Tests ~outstanding~ done
- [x] Review feedback implemented

## References

JIRA ticket:
* https://getpocket.atlassian.net/browse/BACK-1439

### DIFF
--- a/schema-admin.graphql
+++ b/schema-admin.graphql
@@ -201,7 +201,10 @@ type ApprovedCorpusItem @key(fields: "url") {
     """
     Subquery to get the ScheduledCorpusItems to display to a user for a given date
     """
-    scheduledSurfaceHistory(scheduledSurfaceGuid: ID, limit: Int): [ScheduledCorpusItem!]!
+    scheduledSurfaceHistory(
+        scheduledSurfaceGuid: ID
+        limit: Int
+    ): [ScheduledCorpusItem!]!
 }
 
 """
@@ -355,7 +358,7 @@ type ScheduledCorpusItem {
     """
     The associated Approved Item.
     """
-    approvedItem: ApprovedCorpusItem
+    approvedItem: ApprovedCorpusItem!
     """
     The GUID of this scheduledSurface to which this item is scheduled. Example: 'NEW_TAB_EN_US'.
     """
--- a/schema-admin.graphql
+++ b/schema-admin.graphql
@@ -199,12 +199,10 @@ type ApprovedCorpusItem @key(fields: "url") {
     """
     updatedBy: String
     """
-    Subquery to get the ScheduledCorpusItems to display to a user for a given date
+    Subquery to get the log of scheduled entries to display for a given Approved Item, most recent first.
     """
-    scheduledSurfaceHistory(
-        scheduledSurfaceGuid: ID
-        limit: Int
-    ): [ScheduledCorpusItem!]!
+    scheduledSurfaceHistory(filters: ApprovedCorpusItemScheduledSurfaceHistoryFilters):
+        [ApprovedCorpusItemScheduledSurfaceHistory!]!
 }
 
 """
@@ -417,6 +415,45 @@ input ApprovedCorpusItemFilter {
     """
     language: CorpusLanguage
 }
+
+type ApprovedCorpusItemScheduledSurfaceHistory {
+    """
+    An alternative primary key in UUID format that is generated on creation.
+    Note: this is the external ID of the scheduled entry, not the approved item.
+    """
+    externalId: ID!
+    """
+    A single sign-on user identifier of the user who created this entry.
+    """
+    createdBy: String!
+    """
+    The date the associated Approved Item is scheduled to appear on a Scheduled Surface.
+    This date is relative to the time zone of the Scheduled Surface. Format: YYYY-MM-DD.
+    """
+    scheduledDate: Date!
+    """
+    The GUID of the scheduledSurface to which the associated Approved Item is scheduled.
+    Example: 'NEW_TAB_EN_US'.
+    """
+    scheduledSurfaceGuid: ID!
+}
+
+"""
+Available fields for filtering an Approved Item's history of being scheduled onto one or more
+scheduled surfaces.
+"""
+input ApprovedCorpusItemScheduledSurfaceHistoryFilters {
+    """
+    The maximum number of results to be returned. Default: 10.
+    """
+    limit: NonNegativeInt
+    """
+    The scheduled surface the results should be filtered to. Omitting this filter will
+    fetch results from all scheduled surfaces.
+    """
+    scheduledSurfaceGuid: ID
+}
+
 
 """
 Available fields for filtering Rejected Items.
--- a/schema-admin.graphql
+++ b/schema-admin.graphql
@@ -120,13 +120,6 @@ type ScheduledSurface {
 }
 
 """
-Temp type to pacify federation checks
-"""
-type ApprovedCuratedCorpusItem @key(fields: "url") {
-    url: Url!
-}
-
-"""
 A prospective story that has been reviewed by the curators and saved to the corpus.
 """
 type ApprovedCorpusItem @key(fields: "url") {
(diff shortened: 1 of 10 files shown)


schema-admin.graphql
```diff
@@ -119,13 +119,6 @@ type ScheduledSurface {
     prospectTypes: [ProspectType!]!
 }
 
-"""
-Temp type to pacify federation checks
-"""
-type ApprovedCuratedCorpusItem @key(fields: "url") {
-    url: Url!
-}
-
```
nina-py, line 128: Unrelated change - I think this is safe to remove now that Prospect API has been updated to use the shortened `ApprovedCorpusItem` entity instead.

schema-admin.graphql
```diff
@@ -205,6 +198,10 @@ type ApprovedCorpusItem @key(fields: "url") {
     A single sign-on user identifier of the user who last updated this entity. Null on creation.
     """
     updatedBy: String
+    """
+    Subquery to get the ScheduledCorpusItems to display to a user for a given date
+    """
+    scheduledSurfaceHistory(scheduledSurfaceGuid: ID, limit: Int): [ScheduledCorpusItem!]!
```
nina-py, line 204: Note that both filters are optional to future-proof this query. Right now the business case is to fetch this data for a single scheduled surface id + just the one most recent result, but eventually we will need to show on the frontend the entire history for a curated corpus item on demand, and that should show scheduled entries for all surfaces - either all or 6? 12? most recent ones.

schema-admin.graphql
```diff
@@ -358,7 +355,7 @@ type ScheduledCorpusItem {
     """
     The associated Approved Item.
     """
-    approvedItem: ApprovedCorpusItem!
+    approvedItem: ApprovedCorpusItem
```
nina-py, line 358: I have made this optional to reuse this type in the return value of `scheduledSurfaceHistory` subquery. Open to suggestions here! Originally I thought we should set up a very abridged custom type that would only show date + surface guid, but then (see above notes on future-proofing it) decided to return the whole thing. 

But not the approved item! We already have it in the main query and it doesn't make sense to pull it in again.

src/database/queries/ApprovedItem.ts
```diff
@@ -85,6 +85,36 @@ export async function getApprovedItemByUrl(
...
+): Promise<ScheduledItem[]> {
+  const approvedItem = db.approvedItem.findUnique({
+    where: { externalId },
+  });
+
+  return await db.scheduledItem.findMany({
+    take: limit,
+    orderBy: { scheduledDate: 'desc' },
+    where: {
+      approvedItemId: approvedItem.id,
```
nina-py, line 110: This line here breaks everything with a TS error at the moment. It's too late in the day to figure out what I'm doing wrong here.

schema-admin.graphql
Herraj: Wondering if the result returned from this should be a `nullable` value? i.e `[ScheduledCorpusItem!]`, there could be a scenario where it hasn't been scheduled yet for the first time so this shouldn't return any history

src/admin/resolvers/queries/ApprovedItem.ts
```diff
@@ -76,3 +77,53 @@ export async function getApprovedItemByUrl(
...
+  // the entire list of all the occasions this approved item was scheduled
+  // on all surfaces, in descending order (most recently scheduled first).
+  //
+  // Limiting it to a single surface means we'll only retrieve data for
+  // one scheduled surface (useful for when you only need the one to filter out
+  // prospects on the Prospecting page).
+  //
+  // Limiting the number of results to one means only the most recent result
+  // will be returned.
+  const { scheduledSurfaceGuid, limit } = args;
```
kschelonka, line 103: It seems unlikely that there would be a ton of entries to fetch, but I would recommend putting a default limit in place.

src/admin/resolvers/queries/ApprovedItem.ts
```diff
@@ -76,3 +77,53 @@ export async function getApprovedItemByUrl(
...
+
+  // Check if the scheduled surface supplied is valid
+  const surface = ScheduledSurfaces.find((surface) => {
+    return surface.guid === scheduledSurfaceGuid;
+  });
+  if (!surface) {
```
kschelonka, line 109: IMO you could skip this check and just rely on no results being returned for the `where` condition

src/database/queries/ApprovedItem.ts
Herraj: We just had to await the first db call when fetching the `approvedItem` and then do a non-null assertion, you were almost there!

src/admin/resolvers/queries/ApprovedItem.ts
```diff
@@ -76,3 +77,53 @@ export async function getApprovedItemByUrl(
...
+    throw new UserInputError(
+      `Could not find Scheduled Surface with id of "${scheduledSurfaceGuid}".`
+    );
+  }
+
+  // Check if the limit specified passes basic sanity checks
+  if (limit.isNaN() || limit < 1) {
```
kschelonka, line 116: wish we could have finer-grained argument validation on the graphql side easily... almost all numeric args should be > 0

schema-admin.graphql
kschelonka: I don't recommend making this nullable just based off the parent context. Better to use a custom type, return it as expected (and put in reasonable query depth limits to prevent weird circular resolvers), and use dataloaders to cache.

src/database/queries/ApprovedItem.ts
kschelonka: Can you not do joins with prisma?

schema-admin.graphql
bassrock: Agreed @kschelonka we should keep this required given the context and put in query depth limits

schema-admin.graphql
nina-py: Agree these would be needed if we left the original `ScheduledCorpusItem` as a return type in place, but after a discussion with @Herraj and also feedback from @bassrock further down the review comments we decided a custom type that doesn't contain approved items would serve us better here.

nina-py: @Herraj - adding here for posterity as we discussed this in a pairing earlier today - the non-nullable return is for the frontend to have an easier time with return types. Something is always guaranteed to be returned, even if it's just an empty array.

bassrock: @nina-py I think I was unclear. I meant to imply return ScheduledCorpusItem but we can limit the query depth so a user can't infinitely recurse.

src/admin/resolvers/queries/ApprovedItem.ts
nina-py: I could, but I would lose the ability to return an input error to the client, which is something I'd like to do to make the frontend side of things easier during development.

src/admin/resolvers/queries/ApprovedItem.ts
nina-py: Thank you @kschelonka, that jogged my memory as I thought there was something out there to help with this. There's `NonNegativeInt` in `graphql-scalars`, trying it out on this field: https://www.graphql-scalars.dev/docs/scalars/non-negative-int

schema-admin.graphql
bassrock: also just seeing the return type comment for frontend. makes sense

src/admin/resolvers/index.ts
```diff
@@ -1,6 +1,10 @@
-import { DateResolver } from 'graphql-scalars';
+import { DateResolver, NonNegativeIntResolver } from 'graphql-scalars';
```
kschelonka, line 1: 🙌

kschelonka (approved):

PR #688, opened 2022-05-23
fix(authors): [BACK-1442] Add authors to Snowplow output

nina-py:
## Goal

When going through the doc ticket ([#1422](https://getpocket.atlassian.net/browse/BACK-1422)) and referencing some code changes, I realised the initial updates to Curated Corpus API missed adding the authors to the data we send with Snowplow events. This ticket should fix that, and update relevant tests.

- Now sending names of authors as an array to Snowplow along with other data for reviewed corpus items. Updated tests.

## References

JIRA ticket:
* https://getpocket.atlassian.net/browse/BACK-1442

### DIFF
--- a/src/events/types.ts
+++ b/src/events/types.ts
@@ -1,6 +1,13 @@
-import { ApprovedItem, RejectedCuratedCorpusItem } from '@prisma/client';
-import { ScheduledItem, CorpusItem } from '../database/types';
-import { ScheduledItem as ScheduledItemModel } from '@prisma/client';
+import {
+  ApprovedItem as ApprovedItemModel,
+  RejectedCuratedCorpusItem,
+  ScheduledItem as ScheduledItemModel,
+} from '@prisma/client';
+import {
+  ApprovedItemAuthor,
+  ScheduledItem,
+  CorpusItem,
+} from '../database/types';
 
 export enum ReviewedCorpusItemEventType {
   ADD_ITEM = 'ADD_ITEM',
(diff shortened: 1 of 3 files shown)


src/events/types.ts
```diff
@@ -30,6 +37,10 @@ export type BaseEventData = {
   version: string; // semver (e.g. 1.2.33)
 };
 
+export type ApprovedItem = ApprovedItemModel & {
+  authors?: ApprovedItemAuthor[];
+};
```
jpetto, line 42: what do you think about using this definition from `./src/database/types.ts` instead of declaring again here?

jpetto (changes requested):
hoping we can re-use a type def instead of declaring again, but otherwise looks good

src/events/types.ts
nina-py: Good point, thank you @jpetto! I reused the authors from there but somehow overlooked the approved item of the right shape was also already there.

src/events/types.ts
```diff
@@ -50,7 +52,7 @@ export type ScheduledItemEventBusPayload = BaseEventBusPayload &
   Pick<ScheduledItemModel, 'createdBy' | 'scheduledSurfaceGuid'> &
   Pick<
     ApprovedItem,
-    'topic' | 'isSyndicated' | keyof Omit<CorpusItem, 'id' | 'authors'>
+    'topic' | 'isSyndicated' | keyof Omit<CorpusItem, 'id'>
```
jpetto, line 55: this is some deep typescripting 😆 . can you explain this reverted change to me?

jpetto (approved):
lgtm! (but i'd also love a quick explainer on one part...)

src/events/types.ts
nina-py: @jpetto, this went in in the initial PR that added schema changes and updated all the resolvers. I think what happened is that this was added to fix a TS error temporarily to let the rest of the code compile but then this part of the code was forgotten about and so this change went in unnoticed.

PR #689, opened 2022-05-23
fix(authors): Update tests for public query

nina-py:
## Goal

This is a _verrry_ small fix to update integration tests for the public query - adding in authors and expecting to receive some data for them.


jpetto (approved):
nice! love small durability PRs 😄

PR #690, opened 2022-05-27
feat(add topic): add topic to public CorpusItem entity

jpetto:
## Goal

add topic to public `CorpusItem` entity in preparation of need by recommendations/data products.

- update chai statements to use our preferred `exists` (but only in the one file i touched i'm not trying to do it all here)

- `topic` being a string in our graphql schema and not an enum. i personally think this is the right decision at this time, but would like to know if anyone disagrees.

JIRA ticket:
* https://getpocket.atlassian.net/browse/BACK-1452

### DIFF
--- a/src/database/queries/ScheduledItem.ts
+++ b/src/database/queries/ScheduledItem.ts
@@ -129,6 +129,14 @@ export async function getItemsForScheduledSurface(
         language: scheduledItem.approvedItem.language,
         publisher: scheduledItem.approvedItem.publisher,
         imageUrl: scheduledItem.approvedItem.imageUrl,
+        // so the type definition in /src/database/types has topic as optional,
+        // which typescript resolves as `string | undefined`. however, if the
+        // topic is missing in the db, prisma returns `null` - hence the
+        // nullish coalescing operator below.
+        //
+        // i wonder why typescript won't accept both. is there some deep dark
+        // JS reason? or is it just better practice?
+        topic: scheduledItem.approvedItem.topic ?? undefined,
       },
     };
     return item;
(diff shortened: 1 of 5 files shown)


src/database/queries/ScheduledItem.ts
```diff
@@ -129,6 +129,14 @@ export async function getItemsForScheduledSurface(
         language: scheduledItem.approvedItem.language,
         publisher: scheduledItem.approvedItem.publisher,
         imageUrl: scheduledItem.approvedItem.imageUrl,
+        // so the type definition in /src/database/types has topic as optional,
+        // which typescript resolves as `string | undefined`. however, if the
+        // topic is missing in the db, prisma returns `null` - hence the
+        // nullish coalescing operator below.
+        //
+        // i wonder why typescript won't accept both. is there some deep dark
+        // JS reason? or is it just better practice?
```
nina-py, line 138: +1 vote for the "deep dark JS reason" here! :) We could set the type in `/src/database/types` to `topic: string | null | undefined`, but I'm not sure if it would create more problems than it solved!

src/public/resolvers/queries/ScheduledSurfaceItem.integration.ts
```diff
@@ -71,8 +84,8 @@ describe('queries: ScheduledCuratedCorpusItem', () => {
       });
 
       // Good to check this here before we get into actual return values
-      expect(result.errors).to.be.undefined;
-      expect(result.data).not.to.be.null;
+      expect(result.errors).not.to.exist;
+      expect(result.data).to.exist;
```
nina-py, line 88: I like this update!

nina-py (approved):

PR #691, opened 2022-06-03
feat(authors): [BACK-1475] Create a custom mutation to add authors to an approved item

nina-py:
## Goal

Facilitate backfilling author information in the curated corpus. Initially we thought we could use the `updateApprovedCorpusItem` mutation, but that one requires `topic` to be set, which is something that many `DE` items don't yet have. So a custom mutation is needed, and it might as well take in only the data it actually needs rather than all editable fields of a curated item.

- [x] Added a new mutation, `updateApprovedCorpusItemAuthors`, that takes in `externalId` and `authors` object, then updates the approved corpus item.
- [x] Added tests.

Note that a whole bunch of + X lines there, -Y lines here is me splitting out auth-related integration tests for approved items into their own file. 

## References

JIRA ticket:
* https://getpocket.atlassian.net/browse/BACK-1475

### DIFF
--- a/src/admin/resolvers/mutations/ApprovedItem.auth.integration.ts
+++ b/src/admin/resolvers/mutations/ApprovedItem.auth.integration.ts
@@ -1,0 +1,592 @@
+import { CuratedCorpusEventEmitter } from '../../../events/curatedCorpusEventEmitter';
+import {
+  ApprovedItem,
+  ApprovedItemAuthor,
+  CreateApprovedItemInput,
+  RejectApprovedItemInput,
+  UpdateApprovedItemAuthorsInput,
+  UpdateApprovedItemInput,
+} from '../../../database/types';
+import { CuratedStatus } from '@prisma/client';
+import {
+  ACCESS_DENIED_ERROR,
+  CorpusItemSource,
+  MozillaAccessGroup,
+  Topics,
+} from '../../../shared/types';
+import {
+  clearDb,
+  createApprovedItemHelper,
+  getServerWithMockedHeaders,
+} from '../../../test/helpers';
+import { db, getServer } from '../../../test/admin-server';
+import sinon from 'sinon';
+import { ReviewedCorpusItemEventType } from '../../../events/types';
+import {
+  CREATE_APPROVED_ITEM,
+  REJECT_APPROVED_ITEM,
+  UPDATE_APPROVED_ITEM,
+  UPDATE_APPROVED_ITEM_AUTHORS,
+  UPLOAD_APPROVED_ITEM_IMAGE,
+} from './sample-mutations.gql';
+import { expect } from 'chai';
+import { createReadStream, unlinkSync, writeFileSync } from 'fs';
+import { Upload } from 'graphql-upload';
+
+describe('mutations: ApprovedItem - authentication checks', () => {
+  const eventEmitter = new CuratedCorpusEventEmitter();
+
+  // a standard set of inputs for this mutation
+  const input: CreateApprovedItemInput = {
+    prospectId: '123-abc',
+    title: 'Find Out How I Cured My Docker In 2 Days',
+    url: 'https://test.com/docker',
+    excerpt: 'A short summary of what this story is about',
+    authors: [
+      { name: 'Mark Twain', sortOrder: 1 },
+      { name: 'Jane Austen', sortOrder: 2 },
+    ],
+    status: CuratedStatus.CORPUS,
+    imageUrl: 'https://test.com/image.png',
+    language: 'DE',
+    publisher: 'Convective Cloud',
+    topic: Topics.TECHNOLOGY,
+    source: CorpusItemSource.PROSPECT,
+    isCollection: false,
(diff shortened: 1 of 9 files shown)


src/admin/resolvers/mutations/ApprovedItem.ts
```diff
@@ -178,6 +179,59 @@ export async function updateApprovedItem(
...
+  // to fetch the entire object.
+  const existingItem = await getApprovedItemByExternalId(
+    context.db,
+    data.externalId
+  );
+
+  // Remove the old author(s) from the DB records before we run the update function
+  // Note that we don't expect any authors to be present for any items this mutation
+  // will be run for, but it's a good idea to be thorough anyway.
+  if (existingItem) {
```
jpetto, line 210: if there is no `existingItem`, should we halt processing and throw an error? based on the data set we're working with, i can't see this ever happening, BUT if we checking for it anyway, we may as well add in an `else`, yeah?

src/admin/resolvers/mutations/ApprovedItem.integration.ts
```diff
@@ -608,6 +516,73 @@ describe('mutations: ApprovedItem', () => {
...
+        { name: 'Author Two', sortOrder: 2 },
+      ];
+
+      input = {
+        externalId: item.externalId,
+        authors,
+      };
+    });
+
+    it('should succeed if user has full access', async () => {
```
jpetto, line 542: nit - test name could probably be better (as this isn't an auth test)?

src/admin/resolvers/mutations/ApprovedItem.auth.integration.ts
```diff
@@ -0,0 +1,592 @@
...
+      // 2 - Event has the right type.
+      expect(await eventTracker.getCall(0).args[0].eventType).to.equal(
+        ReviewedCorpusItemEventType.ADD_ITEM
+      );
+      // 3- Event has the right entity passed to it.
+      expect(
+        await eventTracker.getCall(0).args[0].reviewedCorpusItem.externalId
+      ).to.equal(result.data?.createApprovedCorpusItem.externalId);
+
+      await server.stop();
```
jpetto, line 104: nit - we can get rid of all the `server.start()` and `server.stop()` calls.

src/admin/resolvers/mutations/ApprovedItem.auth.integration.ts
```diff
@@ -0,0 +1,592 @@
...
+      // 1 - Event was fired once!
+      expect(eventTracker.callCount).to.equal(1);
+      // 2 - Event has the right type.
+      expect(await eventTracker.getCall(0).args[0].eventType).to.equal(
+        ReviewedCorpusItemEventType.UPDATE_ITEM
+      );
+      // 3- Event has the right entity passed to it.
+      expect(
+        await eventTracker.getCall(0).args[0].reviewedCorpusItem.externalId
+      ).to.equal(data?.updateApprovedCorpusItemAuthors.externalId);
```
jpetto, line 380: nit - can we keep the auth tests focused on auth stuff only? we are testing the data in the non-auth tests already, no need to duplicate here.

jpetto (changes requested):
this largely looks great. just a few nits for cleanup.

src/admin/resolvers/mutations/ApprovedItem.ts
nina-py: I'm not sure why it was set up this way (I think I wrote this??? :)), but I bet it was a TypeScript limitation, whereby Prisma could return a `null` for an approved item potentially and TS was not having it. Removing the brackets and using optional chaining instead works, too. I don't think there's a need for an `if` statement here.

src/admin/resolvers/mutations/ApprovedItem.integration.ts
nina-py: True! Should have updated test names when splitting them up.

src/admin/resolvers/mutations/ApprovedItem.auth.integration.ts
nina-py: Awesome! They were put in initially because tests would hang on CircleCI, resulting in a failed check.

src/admin/resolvers/mutations/ApprovedItem.auth.integration.ts
nina-py: Done!

src/database/mutations/ApprovedItem.ts
```diff
@@ -76,7 +77,41 @@ export async function updateApprovedItem(
...
+ * @param username
+ */
+export async function updateApprovedItemAuthors(
+  db: PrismaClient,
+  data: UpdateApprovedItemAuthorsInput,
+  username: string
+): Promise<ApprovedItem> {
+  if (!data.externalId) {
+    throw new UserInputError('externalId must be provided.');
+  }
```
Herraj, line 108: Good check!

Herraj (approved):

PR #692, opened 2022-06-03
feat(resolvers): [HS-32] Resolve CorpusItem

mmiermans:
## Goal
Allow us to resolve Corpus Items, which is needed for the [setupMomentSlate](https://studio.apollographql.com/graph/pocket-client-api/schema/reference?variant=current#setupMomentSlate). This query will be used as part of an experiment in a couple weeks to show new Web users stories during onboarding.

Is there a better place for converting ApprovedItem to CorpusItem than utils.ts?

## References

JIRA ticket:
* https://getpocket.atlassian.net/browse/HS-32

Documentation:
* [Spec PR #144](https://github.com/Pocket/spec/pull/144)

Apollo queries:
* [setupMomentSlate query with only an _id_ field](https://studio.apollographql.com/graph/pocket-client-api/explorer?explorerURLState=N4IgJg9gxgrgtgUwHYBcQC4QEcYIE4CeABAMoIowAOAshIqiQDYCGKCRwAOkkUQM7kqteiiat2XHryJ4EUOvTCsAlhCR8O3adPl5KMPgEk2cTVO29lYLRYC%2BN3valPbIW0A&variant=current)
* [setupMomentSlate query that errors because a title is requested](https://studio.apollographql.com/graph/pocket-client-api/explorer?explorerURLState=N4IgJg9gxgrgtgUwHYBcQC4QEcYIE4CeABAMoIowAOAshIqiQDYCGKCRwAOkkUQM7kqteiiat2XHryJ4EUOvTCsAlhCR8O3adPl5KMPgEk2cTVO29lYLRd4plKRghvaAvi6LupX1yFdA&variant=current)

Sentry error:
* https://sentry.io/organizations/pocket/issues/3319008876/?project=5938284&query=is%3Aunresolved

### DIFF
--- a/src/shared/utils.ts
+++ b/src/shared/utils.ts
@@ -1,6 +1,9 @@
 import { ScheduledSurface, ScheduledSurfaces } from './types';
-import {ApprovedItem, ApprovedItemAuthor, CorpusItem, ScheduledSurfaceItem} from "../database/types";
-import {DateTime} from "luxon";
+import {
+  ApprovedItem,
+  ApprovedItemAuthor,
+  CorpusItem,
+} from '../database/types';
 
 /**
  * Generate an integer Epoch time from a JavaScript Date object.
--- a/src/shared/utils.ts
+++ b/src/shared/utils.ts
@@ -1,4 +1,6 @@
 import { ScheduledSurface, ScheduledSurfaces } from './types';
+import {ApprovedItem, ApprovedItemAuthor, CorpusItem, ScheduledSurfaceItem} from "../database/types";
+import {DateTime} from "luxon";
 
 /**
  * Generate an integer Epoch time from a JavaScript Date object.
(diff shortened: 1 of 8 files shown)


src/shared/utils.ts
```diff
@@ -73,4 +78,27 @@ export const getScheduledSurfaceByGuid = (
...
+
+export const getCorpusItemFromApprovedItem = (
+  approvedItem: ApprovedItem
+): CorpusItem => {
+  return {
+    id: approvedItem.externalId,
+    url: approvedItem.url,
+    title: approvedItem.title,
+    excerpt: approvedItem.excerpt,
+    authors: approvedItem.authors as ApprovedItemAuthor[],
```
mmiermans, line 90: Note: we type cast this, because `authors` is optional for `ApprovedItem`, but required for `CorpusItem`. It's probably also required on `scheduledItem.approvedItem`?

kkyeboah (approved):

src/shared/utils.ts
kschelonka: Good inheritance question...

nina-py: `authors` was made optional because most corpus items don't have one yet - this is a change that was brought in to enable scheduling of Pocket Hits correctly - authors are only used there at the moment. We're working on backfilling authors data for all curated items. Once that is done, we'll be able to make changes both to the admin schema and the underlying database schema.

nina-py (approved):

src/shared/utils.ts
jpetto: as is, because not all approved items have authors yet, i think this will error out depending on the database record returned. are we okay with that? or do we want to make [`authors` on `CorpusItem`](https://github.com/Pocket/curated-corpus-api/blob/main/src/database/types.ts#L158) optional until the backfill completes?

PR #830, opened 2023-01-25
feat(testing-devex): testing developer experience improvements, fixes

hyperparabolic:
This changeset tries to make some improvements to local testing developer experience.

- adapted jest test setup to consume `local.env`, prioritizing already set environment variables
  - de-duplicates config, nothing to get out of sync
  - prefers externally set environment variables, so container behavior is unchanged (environment is pre-set there)
  - keeps config local to the workspace
- docs touchup around testing
- timezone bugs - Date objects returned by prisma are in the system local timezone. All date conversions to strings are off by one if your timezone is a negative offset from UTC. Specify UTC timezone for all Date <-> String conversions.
  - this fixes non-container testing in the Americas, and ensures correct behavior if we somehow deploy in a non-UTC timezone.

## Goal
What changed? What is the business/product goal?

Ironed out a few pain points for local testing while working through the apollo v4 migration. Wanted to get this in a separate PR to avoid any ambiguities related to the migration.

Improves the experience of testing outside containers, and reduces the odds of "works on my machine".

- What tools do people like for interacting with dates? These types of problems are hard enough to keep straight without mixing luxon and native date functions. It would be nice to consolidate.

## Deployment steps
N/A


hyperparabolic:
Something failed while cleaning up after tests in circleci. It looks like tests are running in `cimg/node:current`, which is node 19 and prisma is unhappy about that? All tests passed before it failed? Definitely some weirdness to clean up here...

hyperparabolic:
Current blockers:
- I cannot rotate the snowplow api key for `Verify Snowplow Schema Dependencies`
- Prisma is still unhappy on a node 16 version.  I think this might be related postinstall client setup issues?

hyperparabolic:
Commits are a mess here.  I'll squash the noise away when it's time to merge.

README.md
```diff
@@ -128,17 +128,15 @@ docker compose exec app npm run test-integration:watch
 
 If you'd like to be able to run and debug integration tests directly in your IDE, run the following command (note that it may prompt you for your `sudo` password to modify your `/etc/hosts` file):
 
-```bash
-npm run test-setup
-```
-
-Thereafter, you can use
+Tests also run fine outside the container if you want to attach a debugger or run individual tests. Connection strings are swapped to use `localhost` instead of docker host names for this case (see `./src/test/setup/jest.setup.js`). There shouldn't be much difference between this and running in a container, but tests **MUST** pass in the container to make it through CI.
```
sri-kirsh, line 131: 👍

src/database/queries/ApprovedItem.ts
```diff
@@ -206,9 +206,9 @@ export async function getScheduledSurfaceHistory(
     return {
       ...item,
       // Format the scheduled date to YYYY-MM-DD format
-      scheduledDate: DateTime.fromJSDate(item.scheduledDate).toFormat(
-        'yyyy-MM-dd'
-      ),
+      scheduledDate: DateTime.fromJSDate(item.scheduledDate, {
+        zone: 'utc',
```
sri-kirsh, line 210: trying to understand why we are doing this - if we set the timezone in the sql docker, then it by default considers all date in utc format right?

sri-kirsh (approved):

src/database/queries/ApprovedItem.ts
hyperparabolic: This is bug prevention for cases where the system timezone is not UTC. This keeps tests working well on developer machines outside containers, and is just extra insurance that we do not have off by one date issues if somehow this accidentally gets deployed in a system with a non-utc timezone.

hyperparabolic: There was more conversation around this in slack, putting a quick reproduction snippet here showing the exact thing we're trying to prevent here.  In a node REPL on a system with a negative UTC offset:
```
> const scheduledDate = new Date('2050-01-01T00:00:00.000Z')
undefined
> scheduledDate.toLocaleString('en-US', { dateStyle: 'medium' })
'Dec 31, 2049'
> scheduledDate.toLocaleString('en-US', { dateStyle: 'medium', timeZone: 'UTC' })
'Jan 1, 2050'
```

The luxon DateTime behaves identically.

hyperparabolic: Both dates say `'Jan 1, 2050'` for systems that are UTC or have a positive UTC offset as well.

PR #832, opened 2023-01-27
feat(prospecting): [INFRA-965] Add new prospecting source (TOP_SAVED) to New Tab (US)

nina-py:
## Goal

Add a new prospecting source to the New Tab (US) surface as requested by the data products team. This source is already available for the Pocket Hits (US) surface. 

Similar to the Pocket Hits surfaces, I placed this source on top of the list in the API code. However, on the frontend, the sources are listed alphabetically, so this new source turns up on the list last (see screenshot below).

Additionally, the Snowplow schema check is failing for all PRs in this repository (`Invalid api key`). This won't fix it, but I thought I'd update the schema versions all the same - two out of three were well behind the current version.

topsaved_frontend

## References

JIRA ticket:
* https://getpocket.atlassian.net/browse/INFRA-965

### DIFF
--- a/.github/workflows/snowplow/schemas.json
+++ b/.github/workflows/snowplow/schemas.json
@@ -6,13 +6,13 @@
         "vendor": "com.pocket",
         "name": "object_update",
         "format": "jsonschema",
-        "version": "1-0-5"
+        "version": "1-0-9"
       },
       {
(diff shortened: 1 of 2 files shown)


jpetto (approved):
code is all good - gotta dig into this snowplow check issue...

PR #833, opened 2023-01-27
Updated CI Setup

cmharlow:
Ticket: None, came up in context of CI fails on #830 

What this PR does:
* updates caching strategy in CircleCI to keep cache aligned with changes in packages & prisma config changes
* changes the CI steps for testing, builds to run on the same base image that the application itself is built upon & run within - to avoid differences in deployed environments & CI of OS-level requirements, like openssl versions installed

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,6 +4,12 @@ orbs:
   aws-ecs: circleci/aws-ecs@2.2.1
   pocket: pocket/circleci-orbs@1.2.8
 
+parameters:
+  cache-version:
+    default: node-v2
+    description: "Version prefix for the cache name, primarily for node_modules caching in this pipeline."
+    type: string
+
 # Workflow shortcuts
 not_main: &not_main
   filters:
@@ -46,7 +52,7 @@ only_dev: &only_dev
 jobs:
   apollo:
     docker:
-      - image: cimg/node:current@sha256:60048477384ccd5c594740396d743d2296e0dd637eafa4a80dc51612fd779b41
+      - image: node:16@sha256:496085d0cd8edf49c97ec86ac11bed85b6f965d9b29885887bb0edba26549b07
     steps:
       - checkout
       - run:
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -188,17 +192,15 @@ jobs:
...
       - restore_cache:
           keys:
-            # when lock file changes, use increasingly general patterns to restore cache
-            # If this gets our of sync, you can increment the version (vX).  Just be
-            # sure to match that version when saving the cache as well.
-            - node-v1-{{ checksum "package-lock.json" }}
-            - node-v1-
+            - <<pipeline.parameters.cache-version>>-{{ checksum "package-lock.json" }}-{{ checksum "prisma/schema.prisma" }}
+            - <<pipeline.parameters.cache-version>>-{{ checksum "package-lock.json" }}
+            - <<pipeline.parameters.cache-version>>-
```
hyperparabolic, line 197: Most of this looks good, but I did want to ask just to be positive, do we really want these more general cache keys?

I feel like this can potentially suppress errors here until further in the deployment process. The `db:generate-client` script doesn't do any cleanup, so it only makes additive changes to `node_modules/.prisma/client`.  If we ever remove a `binaryTarget`, or if prisma changes its file structure, these more generic cache keys will prevent those files from ever being cleaned up.  We could also change the `db:generate-client` script to remove the `node_modules/.prisma` directory as well to throw out another option.

hyperparabolic (commented):
Thoughts?

The rest of this looks great.  I'm also following up SRE to take one last shot at rotating snowplow API keys to hopefully get that job working.

.circleci/config.yml
cmharlow: I see no reason to keep the more general keys around, this is mostly just copy-pasta from my own testing. What you're saying makes sense and I can update this PR to match.

hyperparabolic (approved):
Thanks for digging into this with me! Looks good. The snowplow validation failure is due to key rotation issues. It should clear up without any code changes here.

.circleci/scripts/setup_db.sh
```diff
@@ -3,7 +3,7 @@
 
 echo "Setting up database"
 
-sudo apt-get update && sudo apt-get install -y mysql-client
+sudo apt-get update && sudo apt-get install -y default-mysql-client
```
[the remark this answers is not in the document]
cmharlow: os changed.

hyperparabolic (approved):
Re-approved!

PR #834, opened 2023-02-01
feat(saved): allow access to curated info on list

collectedmind:
## Goal
Allow curated info to be available to saved items so we may create a consistent UX across surfaces.

- Create a connection for saved item by url

Everything!

## Implementation Decisions
Curated info is not available to saved items.  This creates opportunity for a confusion user experience because an item they save from one of our surfaces may end up being representing in an entirely different way once it gets to the save surface.  As we move down the path of blurring the lines between these surfaces with things like lists, it is important that we have the most relevant and consistent data available.

Here is a rather visceral example.

### DIFF
--- a/src/public/resolvers/index.ts
+++ b/src/public/resolvers/index.ts
@@ -1,7 +1,7 @@
 import { DateResolver } from 'graphql-scalars';
 import { getScheduledSurface } from './queries/ScheduledSurface';
 import { getItemsForScheduledSurface } from './queries/ScheduledSurfaceItem';
-import { getCorpusItem } from './queries/CorpusItem';
+import { getCorpusItem, getSavedCorpusItem } from './queries/CorpusItem';
 
 export const resolvers = {
   // The Date resolver enforces the date to be in the YYYY-MM-DD format.
@@ -16,7 +16,7 @@ export const resolvers = {
   },
   // Allow the `SavedItem` to resolve the corpus item
   SavedItem: {
-    corpusItem: getCorpusItem,
+    corpusItem: getSavedCorpusItem,
   },
   Query: {
     // Gets the metadata for a Scheduled Surface (for example, New Tab).
--- a/src/public/resolvers/index.ts
+++ b/src/public/resolvers/index.ts
@@ -14,6 +14,10 @@ export const resolvers = {
   CorpusItem: {
     __resolveReference: getCorpusItem,
   },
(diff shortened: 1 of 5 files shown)


src/public/resolvers/index.ts
```diff
@@ -14,6 +14,10 @@ export const resolvers = {
   CorpusItem: {
     __resolveReference: getCorpusItem,
   },
+  // Allow the `SavedItem` to resolve the corpus item
+  SavedItem: {
+    corpusItem: getSavedCorpusItem,
```
bassrock, line 19: @collectedmind needed to use a separate resolver, because I forgot the reference resolvers (the one above this line) pass in different parameters, so we needed to create a slightly modified one for regular fields.

src/public/resolvers/queries/CorpusItem.ts
```diff
@@ -19,3 +22,19 @@ export async function getCorpusItem(item, { db }): Promise<CorpusItem> {
...
+export async function getSavedCorpusItem(
+  item,
+  args,
+  { db }
+): Promise<CorpusItem> {
+  const { url } = item;
+
+  const approvedItem = await getApprovedItemByUrl(db, url);
+  if (!approvedItem) {
+    console.log(`Could not find Corpus Item with Url of "${url}"`);
```
jpetto, line 35: this is a huge nitpick for this PR, but - given the volume of attempts at connecting a save to a corpusitem, i am afraid this `console.log` is going to flood our logs, making it difficult to find anything else. (i just ran into this with the syndication api migration.)

i'm generally in favor of a good `console.log`, but i'm not sure what purpose it's serving here.

please ignore me if the volume here will be far lower than i'm guessing.

jpetto (dismissed):
approving as everything looks good functionally, but if we can remove that `console.log`, let's.

src/public/resolvers/queries/CorpusItem.ts
Herraj, line 35: Did we mean to keep this in?

Herraj (approved):

src/public/resolvers/queries/CorpusItem.ts
bassrock: Yea that's fair we copied the pattern from syndication where it does the same thing.

Herraj (approved):

PR #839, opened 2023-02-01
feat(deps): [INFRA-895] migrate to apollo v4

hyperparabolic:
## Goal

Migrate to apollo v4

- added ContextManager for the public server, and renamed some interfaces to avoid naming conflicts
- server startup refactoring
  - updating plugin boilerplate to be consistent with other services
  - express and apollo server startup separated to acommodate public / admin split
- integration tests refactored to execute through express server with no header mocking.
  - image upload test has been disabled, see corresponding comment for explanation.

There were some minor breaks related to responses transitioning from undefined to null or vice versa while moving away from server.executeOpration. hunting down these differences 1 by 1 and fixing them was getting time consuming due to differences in chai grammar, so I took a pass on expect statements to try to make them more consistent with the following rules based on the API designs I've generally seen in other internal services and proposals, and trying to make error messages more useful if something starts to fail:

  1. errors are always defined or undefined, but never null.
  2. data should be null for cases where there are errors, or not null if we expect no errors (this is not the case for errors that are returned by apollo without interacting with resolvers).
  3. individual properties either check for `.exists` (not null or undefined), or `not.null`, but not `not.undefined`. Querying for fields that have been removed results in a graphql error. Only modifying the test query and removing a property would ever result in the property being undefined.

This PR requires manual testing in dev for image uploads.

- Do we have any ideas for breaking these PRs into smaller chunks? It'd be easier to stage into chunks if the code still built even when tests fail to build and instead failed during the test step, but without that any PR level validation attached to the PR is useless, and nothing works without a lot of force pushes to staging branches.

## Implementation Decisions
- Automated testing for image uploads don't really work without mocking, and there's not a super clear place to inject that anymore.  Tried to give a thorough explanation here: https://github.com/Pocket/curated-corpus-api/compare/main...migrate-apollo-v4?body=&expand=1#diff-a289dfba9eb59503aba3f9fa3c1f09353e06ccdf4be07a91f87132c814c3a24fR942

## Deployment steps
- [X] Validate this PR in dev (currently deployed there, and I have tested adding prospects to the corpus, rejecting them, and image upload).

## References

JIRA ticket:
* [Link to JIRA ticket](https://getpocket.atlassian.net/browse/INFRA-895?atlOrigin=eyJpIjoiNGZlNzg1M2U2N2UzNGExNzlhODIxNTUxNjFiMDQwNGYiLCJwIjoiaiJ9)


src/admin/resolvers/mutations/ApprovedItem.integration.ts
```diff
@@ -832,7 +937,38 @@ describe('mutations: ApprovedItem', () => {
       unlinkSync(testFilePath);
     });
 
-    it('it should execute the mutation without errors and return the s3 location url', async () => {
+    it.skip('it should execute the mutation without errors and return the s3 location url', async () => {
```
kschelonka, line 940: 👍

kschelonka (dismissed):
🎠

kschelonka (approved):

PR #840, opened 2023-02-02
fix(target): updating resolver for target type

bassrock:
## Goal

Fix https://sentry.io/organizations/pocket/issues/3911297928/?project=5938284&query=is%3Aunresolved&referrer=issue-stream

Turns out that the main type (in this case CorpusTarget) comes into resolveType, and we are already defining the CorpusTarget slug elsewhere, we just needed to pass the type name with it.

### DIFF
--- a/src/shared/utils.ts
+++ b/src/shared/utils.ts
@@ -108,6 +108,7 @@ export const getCorpusItemFromApprovedItem = (
     topic: approvedItem.topic ?? undefined,
     target: target?.key && {
       slug: target.key,
+      __typename: target.type,
     },
   };
 };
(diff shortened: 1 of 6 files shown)


src/shared/utils.ts
```diff
@@ -108,6 +108,7 @@ export const getCorpusItemFromApprovedItem = (
     topic: approvedItem.topic ?? undefined,
     target: target?.key && {
       slug: target.key,
+      __typename: target.type,
```
bassrock, line 111: This was all we needed!

kschelonka (approved):

PR #842, opened 2023-02-02
feat(reference): adding a url based reference resolver for CorpusItem

bassrock:
## Goal

Setting up a url reference resolver that will be used to allow CorpusItem to be resolved by it's URL. This will be used on https://github.com/Pocket/list-api/pull/630 which will let clients know if the corpus item has been saved to the user's list.

### DIFF
--- a/src/public/resolvers/queries/CorpusItem.ts
+++ b/src/public/resolvers/queries/CorpusItem.ts
@@ -7,17 +7,18 @@ import { getCorpusItemFromApprovedItem } from '../../../shared/utils';
 import { UserInputError } from 'apollo-server-errors';
 
 /**
- * Pulls in approved corpus items for a given id.
+ * Pulls in approved corpus items for a given id or url.
  *
- * @param item
+ * @param item { id, url }
  * @param db
  */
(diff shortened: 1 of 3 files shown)


src/public/resolvers/queries/CorpusItem.ts
```diff
@@ -7,17 +7,18 @@ import { getCorpusItemFromApprovedItem } from '../../../shared/utils';
...
-export async function getCorpusItem(item, { db }): Promise<CorpusItem> {
-  const { id } = item;
+export async function getCorpusItem({ id, url }, { db }): Promise<CorpusItem> {
+  const approvedItem = id
+    ? await getApprovedItemByExternalId(db, id)
+    : await getApprovedItemByUrl(db, url);
 
-  const approvedItem = await getApprovedItemByExternalId(db, id);
   if (!approvedItem) {
-    throw new UserInputError(`Could not find Corpus Item with ID of "${id}".`);
```
bassrock, line 20: We should not be throwing errors if things are not found in reference resolvers. In this case, it can return null and is not a UserInputError. its just non-existent.

kschelonka (approved):

PR #891, opened 2023-05-26
chore(package upgrades): Migrate to Prisma v.4

nina-py:
## Goal

Update Prisma packages & schema to the latest major version. Upgrade guide is here: https://www.prisma.io/docs/guides/upgrade-guides/upgrading-versions/upgrading-to-prisma-4#client-changes. Luckily, only minor updates to the schema were needed (the explicit `@unique` modifiers). 

- Also fixed deprecation notices from the `faker` package that were adding noise to integration tests


prisma/schema.prisma
```diff
@@ -30,80 +18,67 @@ model ApprovedItemAuthor {
...
   createdBy       String               @db.VarChar(255)
   updatedAt       DateTime             @updatedAt
   updatedBy       String?              @db.VarChar(255)
-
-  // associated models
-  scheduledItems ScheduledItem[]
-
-  // indexes
-  @@unique([externalId])
-  @@unique([url])
```
nina-py, line 63: These have moved to the field declarations after running the upgrade script.

prisma/schema.prisma
```diff
@@ -30,80 +18,67 @@ model ApprovedItemAuthor {
...
+  externalId String   @unique @default(uuid()) @db.VarChar(255)
+  url        String   @unique @db.VarChar(500)
   title      String?  @db.VarChar(255)
-  topic      String?   @db.VarChar(255)
+  topic      String?  @db.VarChar(255)
   language   String?  @db.VarChar(2)
   publisher  String?  @db.VarChar(255)
-  // Can be multiple reasons. For the MVP, Snowplow and the frontend
-  // will be the only consumers of this data.
-  // Will be likely stored either as comma-separated values or JSON.
```
nina-py, line 78: The migration script stripped out all comments - perhaps for the better? As some of them appear to be outdated

prisma/schema.prisma
```diff
@@ -30,80 +18,67 @@ model ApprovedItemAuthor {
...
-  scheduledItems ScheduledItem[]
-
-  // indexes
-  @@unique([externalId])
-  @@unique([url])
+  prospectId      String?              @db.VarChar(255)
+  isTimeSensitive Boolean              @default(false)
+  source          CuratedItemSource?
+  authors         ApprovedItemAuthor[]
+  scheduledItems  ScheduledItem[]
```
nina-py, line 42: The migration script has also moved fields around for... reasons? Should I put them back where they were to reduce the diff?

src/admin/resolvers/mutations/ScheduledItem.ts
```diff
@@ -11,7 +11,7 @@ import {
   AuthenticationError,
   UserInputError,
 } from '@pocket-tools/apollo-utils';
-import { PrismaClientKnownRequestError } from '@prisma/client/runtime';
+import { PrismaClientKnownRequestError } from '@prisma/client/runtime/library';
```
nina-py, line 14: This also generated a "deprecated" notice in the tests

prisma/schema.prisma
sri-kirsh: ya looks like it grouped all fields that non primitive object types together

likewise, nullable fields at the end and non-nullable at start..this format is good, more organized.

sri-kirsh (approved):

PR #908, opened 2023-05-31
feat(synchecks): add basic synchecks for graphql, server

cmharlow:
Jira ticket: https://getpocket.atlassian.net/browse/INFRA-1276

**What this PR does:**
Adds some synthetic checks, using baseline of reusable code from monitoring "project" work, for sake of monitoring curated corpus api better in service of New Tab.

**Why this PR:**
New Tab is ... going live? Doing something important this week <New Tab folks to fill in>. As part of this, they requested that all services New Tab depends on are better supported operationally ASAP. Alas, Curated Corpus API's operational ownership falls currently to Backend.

As part of trying to be at least a little bit more prepared operationally, this PR adds a few synthetic checks to better monitor deployed Curated Corpus API codepaths that are vital for New Tab: the scheduledSurface Query, and something to do with resolving CorpusItem-typed resources. Plus we should be aware if the express server is up or not.

**The synthetics added:**
* check if the express server is up or not just pinging the healthcheck & validating the response body & response code (directly, inside the VPC, not getting messed up if Client API fails for some reason - this is true of all these synthetics);
* POST a scheduledsurface query & parse & validate the graphql response (to make sure that query is working);
* Ensure same scheduledSurface query POST returns resolved items & CorpusItems, validating the graphql response lengths for returned corpusItem.ids is non-null (to make sure the CorpusItem resolution is working);

**Questions about these:**
* ~is this the best way to check the fitness of these Graphql resources & operations per how New Tab uses this service? I'm just guessing at the most obvious paths~; conferred with Nina & Jonathan independently on this and made adjustments accordingly.
* ~it appears that the curated corpus database does not remove items from the ScheduledItem table in the curated corpus database (dev & prod) once the scheduledDate pasts - giving me something to validate resolution on. Please tell me if this is incorrect though, primarily for prod (if dev alerts, it won't page, it'll just be red, so I care a little less).~ Confirmed items in database shouldn't be removed after date passes.
* I'm not validating exact ids returned for scheduledItem table values since I don't trust the sort order will remain consistent. If we trust the order of CorpusItem.ids returned is sorted consistently, we could validate specific id values instead of number of ids returned.

**Next Steps:**

- [x] get this into dev and validate - already done, dev is working fine
- [ ] get this into prod without alerting / paging hooked up & validate - will occur after this PR goes in
- [ ] make a second PR that enables alerting based on these synthetics 
- [ ] update https://getpocket.atlassian.net/wiki/spaces/PE/pages/2561343516/Curated+Corpus+API to match the service doc template, including reliability docs, links to alerting & metrics, etc. - so we can know whats expected of this service going forward

### DIFF
--- a/.aws/src/main.ts
+++ b/.aws/src/main.ts
@@ -15,6 +15,7 @@ import { config } from './config';
 import {
   ApplicationRDSCluster,
   PocketALBApplication,
+  PocketAwsSyntheticChecks,
   PocketECSCodePipeline,
   PocketPagerDuty,
   PocketVPC,
@@ -28,10 +29,9 @@ class CuratedCorpusAPI extends TerraformStack {
     super(scope, name);
 
     new AwsProvider(this, 'aws', { region: 'us-east-1' });
-
-    new PagerdutyProvider(this, 'pagerduty_provider', { token: undefined });
     new LocalProvider(this, 'local_provider');
     new NullProvider(this, 'null_provider');
+    new PagerdutyProvider(this, 'pagerduty_provider', { token: undefined });
 
     new RemoteBackend(this, {
       hostname: 'app.terraform.io',
@@ -39,14 +39,15 @@ class CuratedCorpusAPI extends TerraformStack {
       workspaces: [{ prefix: `${config.name}-` }],
     });
 
-    const pocketVpc = new PocketVPC(this, 'pocket-vpc');
-    const region = new DataAwsRegion(this, 'region');
     const caller = new DataAwsCallerIdentity(this, 'caller');
+    const region = new DataAwsRegion(this, 'region');
+    const pocketVpc = new PocketVPC(this, 'pocket-vpc');
+    const curatedCorpusPagerduty = this.createPagerDuty();
 
     const pocketApp = this.createPocketAlbApplication({
       rds: this.createRds(pocketVpc),
       s3: this.createS3Bucket(),
-      pagerDuty: this.createPagerDuty(),
+      pagerDuty: curatedCorpusPagerduty,
       secretsManagerKmsAlias: this.getSecretsManagerKmsAlias(),
       snsTopic: this.getCodeDeploySnsTopic(),
       region,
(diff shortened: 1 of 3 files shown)


cmharlow:
Here are some jmespath functions we could investigate using too if we don't like comparing the length of IDs returned for that last synthetic check: https://jmespath.org/proposals/functions.html

nina-py:
Answers:
_-  is this the best way to check the fitness of these Graphql resources & operations per how New Tab uses this service? I'm just guessing at the most obvious paths;_

For New Tab, yes. There are Admin API-side queries that would also benefit from these new synthetic alarms, but they're out of scope for this ticket definitely and probably a lot lower priority overall.

_- it appears that the curated corpus database does not remove items from the ScheduledItem table in the curated corpus database (dev & prod) once the scheduledDate pasts - giving me something to validate resolution on. Please tell me if this is incorrect though, primarily for prod (if dev alerts, it won't page, it'll just be red, so I care a little less)._

Yep, those scheduled entries are meant to be kept forever. 

    _I'm not validating exact ids returned for scheduledItem table values since I don't trust the sort order will remain consistent. If we trust the order of CorpusItem.ids returned is sorted consistently, we could validate specific id values instead of number of ids returned._

The current sort order is by `lastUpdated` if I remember correctly - due to the need to have some sort of predictable ordering for Pocket Hits surfaces. Firefox takes in the list of items for New Tab and ranks it independently of Pocket APIs, so even the `lastUpdated` stuff may not be relied on/permanent.

.aws/src/main.ts
```diff
@@ -28,32 +29,64 @@ class CuratedCorpusAPI extends TerraformStack {
...
+          jmespath: 'data.scheduledSurface.id',
+          response: 'NEW_TAB_EN_US',
+        },
+        {
+          endpoint: config.domain,
+          data: '{"query": "query { scheduledSurface(id: \\"NEW_TAB_EN_US\\") {items(date: \\"2023-05-30\\") {corpusItem {id, url}}}}"}', // New Tab also relies upon corpusItem resolution
+          jmespath:
+            'to_string(length(data.scheduledSurface.items[*].corpusItem.id))',
+          response: `${config.environment === 'Prod' ? '25' : '2'}`,
+        },
```
nina-py, line 77: I would not hardcode the IDs expected here - rather just make sure they are returned and are of the correct type (e.g. integer) if that is possible with `jmespath`. For a thorough check, I'd also make sure _all_ fields are returned without an error - in the past, we've had issues with queries where querying for one specific field would make the whole query die with a 500 error, whereas the same query without that field include was fine.

cmharlow: So per discussion last night with Nina:
1. these aren't hardcoded ids - the existing check is confirming the length of items returned matches what exists for this older date in dev & prod.
2. dev may fail, but we have no guarantees of uptime on dev, hence no alerting on dev, so that's fine - the dev synthetic check primarily exists to confirm the fitness of the check, not the uptime of the dev service.
3. there's a new feature request here for synthetics to do more than issue request confirm response via matching - there's a shape validation request, and possibly a smaller request to do more than matching - have variable comparison operators (so you can do greater than, less than etc). This isn't impossible, just more than synchecks defaults tend towards, so working on some approach for this atm.

Beyond that, its unclear there is a good way to test the fitness / health of specific "entity resolution" in our subgraphs without having known entities the synthetic can faithfully call to confirm. Ideally, we start to roll those out in prod so there are easier checks, but we'll see if the shape validation route can be done without adding much to the syncheck code itself, which should remain as simple as feasible.

cmharlow: Alright, pushed up a change that will fail on this check if a non-empty list of corpusItems show up or any of these shows up without an id field.

we will want to expand eventually to cover more fields on a corpusItem that indicate health (or not) of the service, but need to balance operational health vs data validation / code validation, which are different vectors of alarm.

.aws/src/main.ts
```diff
@@ -28,32 +29,60 @@ class CuratedCorpusAPI extends TerraformStack {
...
+    new PocketAwsSyntheticChecks(this, 'synthetics', {
+      // alarmTopicArn: config.environment === 'Prod' ? curatedCorpusPagerduty.snsCriticalAlarmTopic.arn : '', // this should be improved, empty string recreates updates constantly as is in cdktf
+      alarmTopicArn: '', // remove this line & uncomment above line when we're ready to alert on synchecks
+      environment: process.env.NODE_ENV === 'development' ? 'Dev' : 'Prod', // yes we should use config.environment, but needs more refinment in module
+      prefix: config.prefix,
+      query: [
+        {
+          // New Tab relies upon scheduledSurface query & upon corpusItem resolution
+          endpoint: config.domain,
+          data: '{"query": "query { scheduledSurface(id: \\"NEW_TAB_EN_US\\") {items(date: \\"2023-05-30\\") {corpusItem {id, url}}}}"}',
```
Herraj, line 68: nit: Wondering if we can use ` `` ` here to avoid the ` \\ `?

Herraj (approved):

.aws/src/main.ts
cmharlow: confirmed in dev - seems like we have to keep the \\ / double escapes here because of some of the issues with jsonencode by tf passing this data to aws.

PR #967, opened 2023-07-11
Get description metadata

rolf-moz:
Created a lightweight query endpoint to retrieve the OpenGraph metadata from a 3rd party web site.
This is used for collection admin tools only
https://getpocket.atlassian.net/browse/TML-86

Note that this alternative Parser modification was considered but we are trying to limit changes to the parser
https://getpocket.atlassian.net/wiki/spaces/PE/pages/2988441601/Spec+Add+article+description+to+Item
 
image


schema-admin.graphql
```diff
@@ -877,6 +877,11 @@ type Query {
     Retrieves all ScheduledSurfaces available to the given SSO user. Requires an Authorization header.
     """
     getScheduledSurfacesForUser: [ScheduledSurface!]!
+
+    """
+    Tool to get OG description of URL that is not yet implemented in parser
+    """
+    getOpenGraphFields(url: Url!): OpenGraphFields!
```
Herraj, line 884: We can make the return optional / nullable here. I think the benefit of that is in the client code we can just check `if(getOpenGraphField){...}` instead of `if(getOpenGraphField.description){...}` since according to the gql schema, `description` prop on that type could be `null`.

See comments below!

src/admin/resolvers/queries/OpenGraphFields.ts
```diff
@@ -0,0 +1,17 @@
...
+ */
+export async function getOpenGraphFields(
+  parent,
+  { url }
+): Promise<OpenGraphFields> {
+  const pageMetadata = await getMetaData(url, { maxRedirects: 2, timeout: 5 });
+  return {
+    description: pageMetadata?.description || null,
+  };
+}
```
Herraj, line 17: I think the return type conflicts here a little. According to the function return type we always return a `OpenGraphFields` type. In the types file, that type has the `description` property as required (`string`) but we can potentially return it as `null` here. 

I say we change the function return type to `Promise<OpenGraphFields | null>`. And have our `return` return that object or a `null`.

This will also align with our gql schema above since you have the `description` property on `OpenGraphFields` type optional as well. What do you think 🤔

src/admin/resolvers/queries/OpenGraphFields.ts
```diff
@@ -0,0 +1,23 @@
...
+ */
+export async function getOpenGraphFields(
+  parent,
+  { url }
+): Promise<OpenGraphFields | null> {
+  let pageMetadata = null;
+  try {
+    pageMetadata = await getMetaData(url, { maxRedirects: 2, timeout: 6000 });
+  } catch (ex) {
+    // empty
```
Herraj, line 17: I guess we could just return null here as well.

src/admin/resolvers/queries/OpenGraphFields.integration.ts
```diff
@@ -0,0 +1,156 @@
...
+          variables: {
+            url: serverURL,
+          },
+        });
+      expect(data?.getOpenGraphFields.description).to.equal(
+        sampleDescriptionText
+      );
+    });
+
+    it('missing page', async () => {
```
Herraj, line 108: `should return null for missing page`?

Herraj (approved):
Good tests! 🚀

rolf-moz:
Duplicate PR - other PR with signed commits was just merged

PR #990, opened 2023-08-17
feat(prisma): Bump prisma package version

Herraj:
## Goal
Updating prisma, prisma client and devoxa package versions to be in sync with collections-api in order to prevent connection establishment failures.

Paired with Mathijs and Jesh.

JIRA ticket:
* [MC-34](https://mozilla-hub.atlassian.net/browse/MC-34)


mmiermans (approved):
> 7,634 additions, 7,402 deletions not shown because the diff is too large

Too bad that Node went on a reorganizing spree, and shuffled a bunch of packages around, which makes it harder to see which sub-dependencies got changed.

I looked at the `devoxa/prisma-relay-cursor-connection` and Prima changelog. Don't see any breaking changes in either. Prima does list a bunch of fixes which most releases. Can't tell if any of those apply to us.

I'll defer to you when you want to merge this in, given that some images don't load in the dev environment.

mmiermans (approved):

PR #995, opened 2023-08-30
chore(Database): [MC-61] Increase min capacity from 2 to 8

mmiermans:
## Goal
Increase min capacity from 2 to 8, to see if this resolves the database reliability issues we have been facing since August 15th.

CPU is about 30% on average, whereas before August 14th, it was only about 12% on average.

image

## References

JIRA ticket:
* https://mozilla-hub.atlassian.net/browse/MC-61


mmiermans:
> Curious why you picked 8 specifically?

Pretty arbitrarily.. "Go big or go home."

PR #1000, opened 2023-09-01
Set log level to `debug`

jeshuaborges:
## Goal

To find the underlying invalid foreign key error by enabling all logging details.  

## Implementation Decisions

This is going to create huge Cloudwatch logs and also slow down our response time. This is a temporary debug state. I will flush logs and scale the service back down once we have a single incident to isolate an analyze in the logs.


mmiermans (approved):
I deployed this to Pocket-Dev and confirmed that it works there.

I will be monitoring this deployment on production, given the potential increase in CPU and latency.

PR #1007, opened 2023-11-28
feat(OGExcerpt): [MC-318] Add rss_logistic_recent prospect type to new_tab_en_us

Herraj:
## Goal
Add the new `RSS_LOGISTIC_RECENT` prospect type to `NEW_TAB_EN_US` surface for the new tab recent content experiment.

Will be merged before this [PR](https://github.com/Pocket/prospect-api/pull/572).

JIRA ticket:
* [MC-318](https://mozilla-hub.atlassian.net/browse/MC-318)

[MC-318]: https://mozilla-hub.atlassian.net/browse/MC-318?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/schema-admin.graphql
+++ b/schema-admin.graphql
@@ -81,6 +81,7 @@ enum ProspectType {
     TITLE_URL_MODELED
     DISMISSED
     RSS_LOGISTIC
+    RSS_LOGISTIC_RECENT
 }
 
 """
@@ -970,5 +971,5 @@ type CorpusItem @key(fields: "url") {
 }
 
 type OpenGraphFields {
-  description: String!
+    description: String!
 }
--- a/src/shared/types.ts
+++ b/src/shared/types.ts
@@ -13,6 +13,7 @@ export enum ProspectType {
   TITLE_URL_MODELED = 'TITLE_URL_MODELED',
   DISMISSED = 'DISMISSED',
   RSS_LOGISTIC = 'RSS_LOGISTIC',
+  RSS_LOGISTIC_RECENT = 'RSS_LOGISTIC_RECENT',
 }
 
 export enum MozillaAccessGroup {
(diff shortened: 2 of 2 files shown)


schema-admin.graphql
```diff
@@ -81,6 +81,7 @@ enum ProspectType {
     TITLE_URL_MODELED
     DISMISSED
     RSS_LOGISTIC
+    RSS_LOGISTIC_RECENT
```
efixler, line 84: Question(s): What's really happening / being declared here?

1. wrt what's referenced in the leading comment -- that this isn't used here, but is used somewhere else -- how does that thong know about this?
2. Literally what are these defintions? They look like `consts` but I'm inferring that they are really strings and we just don't require quotes here?

efixler:
lgtm. Who/when/how is going to verify that this is working?

efixler (approved):

jpetto (approved):

schema-admin.graphql
Herraj: So in the typescript code below, we have the prospect types as enums. And in the same file below, we have the scheduled surface objects. 

Each scheduled surface has it's own permission level (which curator can read or write) and it has the list of applicable prospect types.

We pass the prospect type enums to that scheduled surface's array of supported prospect types.

This is updating the enum type in the graphQL schema so that graphQL knows that this is a new and valid prospect type. We chose to use enums to enforce type safety and prevent errors since typing a string could lead to a typo and graphql also can do it's own validation with enums.

Once these changes are in and deployed, I will run a script to re-generate types on the front-end (curation-admin-tools repo) against these new schema changes and those prospect types would be immediately available in the front-end code base as valid types _and_ in the UI drop down without having me to touch any code :)

efixler: Thanks! Just to double check a few things:

- So I'm inferring that these are strings, presumably `.graphql` has no way to reference the TS `ProspectType` enum directly.
- Do we actually permission curators differently? What I'm seeing below looks like we associate prospecting sources with _locales_ but I'm wondering if we actually do restrict _editors_ by locale?

Herraj: Replied in Slack!

